PR #23, opened 2026-08-12
CTO: bounded precision fixes — explicit meal slot authority

ButiMM:
CTO branch for Work Order A. Makes explicit user meal-slot claims authoritative over server-clock inference and adds regression coverage. No production deployment performed.

### DIFF
--- a/script/meal-slot-override-test.ts
+++ b/script/meal-slot-override-test.ts
@@ -1,0 +1,19 @@
+import { explicitMealSlotWins } from "../server/understanding/executor";
+
+type Case = { name: string; input: string | undefined | null; expected: string | undefined };
+
+const cases: Case[] = [
+  { name: "explicit lunch survives a morning clock", input: "lunch", expected: "lunch" },
+  { name: "explicit dinner survives a morning clock", input: "dinner", expected: "dinner" },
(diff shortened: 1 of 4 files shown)


ButiMM:
CTO integration directive — bounded only. 1) Update the two unit assertions that encode the obsolete clock-wins rule: explicit user meal slots must remain authoritative; rewrite those tests to assert explicit lunch/dinner survive a conflicting clock. Do NOT weaken the new executor behavior. 2) Integrate `server/response-scope.ts` into the existing grocery/food-command owner: micro/local modification (e.g. `Can I use eggs instead?`) must return a direct local substitution answer; only explicit full-list/update/rebuild requests may call the full `formatShoppingList()` reconstruction. 3) Add the response-scope regression to `npm test`. 4) Add runtime prompt component instrumentation in `askCoachK()` (no behavior change): log char counts for STATIC_HOT_BRAIN, staticGuide, context, patternSummary, client narrative, SA flags, todayFoodContext, memory, tail, total. 5) Raise the conversational completion ceiling from 160 only for long-form/list requests; preserve the 2–3 sentence conversational cap. Do not touch Pulse, DB migrations, voice, or the Reality harness. Run `npm test` and report. Stop.

PR #83, opened 2026-08-27
The card stops selling food on a day the client closed

ButiMM:
## The customer failure

Live phone trace, 2026-08-26. The client said they were done eating for the day. One turn answered them twice, from the same day-state:

```
text: "Get today's session done."                    <- closure respected
card: "Get a real protein into your next two meals"  <- still selling food
```

## The mechanism (runtime, not static)

Two owners of the same question — "what should they do next":

| owner | rungs | closure guard |
|---|---|---|
| `chooseAction` (`one-action.ts`) | `eat_more`, `protein` | `!s.foodDayClosed` |
| `nextMoveLine` (`macro-card-attach.ts`) | the **same two rungs**, on the card | none |

## The correction

No new prose, no new owner, no new module. `nextMoveLine` already owned four close-out lines for the after-20:00 case, and one of them — *"That's the day. Tomorrow get a real protein in at breakfast"* — is exactly the right coaching for someone who has stopped eating short of protein. So **closure enters the same branch the clock does**, and the four lines below it are untouched:

```ts
if (foodDayClosed || hour >= 20) {
```

The card markers read closure from `held-constraints` — the same durable owner `canonicalDecision` consults. The read fails open to "not closed": a card is a bonus, never a blocker.

`+38 / −3` in production code, one file.

## What is proven — stated precisely

**The card no longer sells food when a closed-day constraint is held**, and the closure reaches it through the real production path. That is the claim.

This does **not** claim that the card and the text choose the same next move in general, and it does not address same-turn closure. Those are separate questions and are untouched here.

## The proof runs through the production marker

The first version of this proof called `nextMoveLine(rows, …, foodDayClosed)` directly. Grok's question 4 killed it: delete both `foodDayClosed: await dayClosedFor(...)` lines from the markers and every assertion stayed green while the client kept reading "eat more" on a day they closed. A test that can be green while the customer is wrong is not evidence.

The graded path is now the real one, end to end:

```
dailyMacroCardMarker / macroCardMarker
  -> dayClosedFor -> readHeldConstraints -> nextMoveLine -> PNG
```

The card's words are rasterised into that PNG and cannot be read back, so the assertion is on **bytes**: render the two cards this day-state can produce, then check which one the marker actually emitted. Byte equality is exact — it pins the specific close-out line, not merely "some different card".

**Both call sites are graded.** Grading only one left the other deletable with the suite still green — the same defect one call site along. The meal-log card is the one that rides every food log a client sends.

Three things are pinned so the bytes mean something: the **clock** (frozen at 19:00 SAST on a fixed date — after 20:00 the branch is taken anyway, so broken wiring would look fixed), the **ledger** (an empty day), and the **closure** (seeded through the raw `pool.query` seam, since a constraint stated at 08:00 cannot otherwise be expressed offline).

## Controls — six, each failing independently

| control | result |
|---|---|
| delete the meal-log marker's wiring | **RED** |
| delete the daily-card marker's wiring | **RED** |
| remove `foodDayClosed \|\|` from the branch | **RED** — reproduces the live trace verbatim |
| drop the `mealCard` pass-through | **RED** |
| over-guard the branch with `if (true)` | **RED** — "An open day lost its card instruction" |
| unpin the clock to 21:00 SAST | **RED** — the pin is load-bearing, not decoration |

Every failure message carries the customer-visible string, e.g.:

```
✗ A held closed-day constraint did not reach the meal-log card: it emitted the
  same image as an open day — card said "Get a real protein into your next two meals"
```

## Suite state

```
✓ tracking-contract-tests
suites: 32/35 green, 1 covered nothing
```

Governors, unchanged from main `d978d00`:

```
```

No budget raised, no suppression, no new module. The PNG renderer is already exercised by `unit-tests` in CI, so this adds no new build dependency.

## Commits

| SHA | scope |
|---|---|
| `63d930c` | production: closure enters the close-out branch |
| `68b1bb8` | test-only: the proof runs through the production marker |

## Carried forward, not silently closed

Cut C (goal-distance → trajectory owner) and the step-raise coaching hole from the same evidence set are untouched, per the locked stage order.

### DIFF
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -301,6 +301,90 @@ const MUST_WRITE: [string, string][] = [
   }
 
   /**
+   * THE CARD MAY NOT CONTRADICT THE DECISION (2026-08-26, live phone trace).
+   *
+   * The client said they were done eating for the day. The text respected it. The card told them
+   * to eat more. Both were produced by the same turn, from the same day-state, by two different
+   * owners of the same question:
+   *
+   *     chooseAction        eat_more and protein rungs, each guarded by !s.foodDayClosed
+   *     nextMoveLine        the SAME two rungs on the card, with no such guard
+   *
+   * Reproduced deterministically on one day-state — fat-loss client, 19:00, protein 60 of 180:
+   *
+   *     text: "Get today's session done."                    (closure respected)
+   *     card: "Get a real protein into your next two meals"  (still selling food)
+   *
+   * The fix adds no prose. A closed day is a finished day, and nextMoveLine already owned four
+   * close-out lines for the after-20:00 case — "That's the day. Tomorrow get a real protein in at
+   * breakfast" is exactly right for someone who stopped eating short of protein. Closure enters
+   * the same branch the clock does.
+   *
+   * GRADED THROUGH THE REAL PATH, not just the pure function: the card markers read closure from
+   * held-constraints, the same durable owner canonicalDecision consults. That read goes through
+   * raw pool.query, so it is seeded here via __KAMLIFE_STUB_PGROWS — the seam that exists because
+   * "the client said at 08:00 they are not training today" could not otherwise be expressed
+   * offline. Without it this test would only prove the parameter works, not that anything passes it.
+   */
+  {
+    const { nextMoveLine, mealCard } = await import("../server/macro-card-attach");
+    const { chooseAction } = await import("../server/one-action");
+    const { readHeldConstraints } = await import("../server/held-constraints");
+
+    const rows = [
+      { label: "Calories", current: 1400, target: 2700 },
+      { label: "Protein", current: 60, target: 180 },
+      { label: "Fat", current: 40, target: 80 },
+      { label: "Carbs", current: 100, target: 300 },
+    ] as any;
+    const dayState = (closed: boolean) => ({
+      goal: "fat_loss", weeksOnProgramme: 5, daysSinceAnyLog: 0, daysSinceWeighIn: 1,
+      loggedToday: true, proteinPct: 60 / 180, caloriePct: 1400 / 2700,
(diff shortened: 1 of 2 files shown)


ButiMM:
## CI verified on the exact head `63d930c`

Run [33051713035](https://github.com/ButiMM/KamLife-Coach/actions/runs/33051713035/job/98448582526), completed 07:57 UTC.

```
suites: 32/35 green, 1 covered nothing in 103s
✓ tracking-contract-tests (10.0s)
✓ production-parity (11.8s)
```

**Baseline-equivalent.** Every number is identical to main `d978d00`. The two RED suites are the pre-existing governors, unchanged — no budget raised, no suppression added, no new module. The `⊘` is the standing normalizer-corpus gap, which needs live credentials and is not touched here.

`tracking-contract-tests` is green **with this cut's three new assertions inside it**, each of which was shown to go RED when its mechanism is removed (details in the description).

---

ButiMM:
## CI verified on the new head `68b1bb8`

Run [33053252544](https://github.com/ButiMM/KamLife-Coach/actions/runs/33053252544/job/98453640098), completed 08:18 UTC.

```
✓ tracking-contract-tests (12.8s)
✓ production-parity (14.1s)
suites: 32/35 green, 1 covered nothing in 113s
```

**Baseline-equivalent.** Every number identical to main `d978d00`. The two RED suites are the pre-existing governors — no budget raised, no suppression, no new module.

The contract suite is green **in CI with the byte-comparison wiring proof inside it**, which was the one environment risk worth naming: the cards are rasterised on the runner, and both sides of every comparison are rendered on the same machine, so the equality holds there as it does locally.

The head this ran against is `68b1bb8`, on these two facts: the check-runs query resolves against the PR head, and the run started 08:15:44 UTC — after the push. The combined-status endpoint returns `403 Resource not accessible by integration` under this integration's permissions, so that is not available as a third confirmation; reporting rather than retrying.

Back to the four questions.

---

PR #84, opened 2026-08-27
A step raise earns the same next move as a first report

ButiMM:
## The customer failure

Traced through `handleMessage` on main `5e38582`, before any change. Same client, same message, same day — three outcomes decided only by what was already stored:

| turn | stored today | reply |
|---|---|---|
| **A** first report (INSERT) | none | `8 500 steps — nice one. 👌`<br><br>`Stand on a scale this morning, before you eat.` |
| **B** raise (UPDATE) | 3 000 | `8 500 steps — nice one. 👌` |
| **C** superseded (no change) | 9 000 | `9 000 steps — nice one. 👌` |

The day moved from 3 000 to 8 500 in **B** exactly as it moved from nothing to 8 500 in **A**. The client got a coach in one and a receipt in the other.

## The mechanism

`logStepsForUser` (`server/handlers/steps.ts`) has three exits and only one of them spoke:

```ts
if (steps > (existing[0].steps ?? 0) || opts?.correction) {
  await db.update(stepLogs).set({ steps })…    // state changed, silence
  return steps;
}
return existing[0].steps ?? steps;             // no change, silence is correct
…
await db.insert(stepLogs).values(…);
turnMutation(`INSERT steps=…`);                // only this one records
```

`closeCoachingTurn` asks `durableDomains(turnMutations())` what the turn changed, got nothing, and stood down. So Law 4 — *a durable write is followed by one next move* — was exempting the commonest shape of step report there is: a client topping up a total they already sent, "5k so far" at noon, "9k" at night.

## The correction

Two lines. No new owner, no new module.

1. The raise branch records its write, the way the INSERT already did.
2. `DURABLE_WRITE` matches it — following **water's** precedent, the existing entry for a domain whose durable write is an UPDATE. The verb stays the anchor, so `"TURN committed steps"` still cannot read as a write.

The no-change branch stays silent **on purpose**, and that is graded: a report the day does not take is not a durable write, and manufacturing a coaching move out of a read is the opposite defect.

## Proof

Graded as a **pair** — the contrast is the evidence. Grading the raise alone would pass on a build where neither turn gets a move.

| control | result |
|---|---|
| delete the mutation from the raise branch | **RED** — reproduces the trace verbatim: `"8 500 steps — nice one. 👌"` |
| revert `DURABLE_WRITE` to `/INSERT steps/i` | **RED** — same failure, proving the pattern half is load-bearing too |
| record the mutation unconditionally | **RED** — the superseded report invents a move: `"Make your next meal a proper protein meal."` |

Both halves of the fix fail independently, and the over-fire direction is guarded.

## Suite state

```
✓ tracking-contract-tests
suites: 32/35 green, 1 covered nothing in 99s
```

Governors, unchanged from main `5e38582`:

```
```

No budget raised, no suppression, no new module.

## Found while tracing — NOT fixed here, and NOT yet proven

`turnAlreadyWrote` (`server/handlers/chat-log.ts:557`) holds a **second** domain→pattern map that still reads `/INSERT steps/i`. It answers a different question — "has a door already written this domain this turn, so a later door should stand down?" — and this change does not alter its behaviour, since the new mutation string does not match its regex.

But it means a step **raise** is invisible to the duplicate-write guard. Whether that is reachable is **unverified**: I have not traced a path where the multi-day dating owner produces an UPDATE and a downstream single-day door then writes again. On this repo static candidates have been ~100% false positives, so it is recorded as a candidate, not a defect, and deliberately not acted on inside this cut.

---

### DIFF
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -657,7 +657,7 @@ const MUST_WRITE: [string, string][] = [
    * week; the ordering control alone asks for `proteinAtTarget`, and it is the only case that
    * depends on which day it runs.
    */
-  async function coachingTurn(message: string, opts?: { seeWrites?: boolean; proteinAtTarget?: boolean; weeksOnProgramme?: number }) {
+  async function coachingTurn(message: string, opts?: { seeWrites?: boolean; proteinAtTarget?: boolean; weeksOnProgramme?: number; stepsAlready?: number }) {
     // BOTH THE COLUMN NAME AND THE SELECT ALIAS. The stub returns raw rows and does not apply
     // drizzle projections, so `db.select({ kcal: mealLogs.kcalInt })` reads undefined and the
     // client silently looked like they had logged NOTHING today — which quietly moved the ladder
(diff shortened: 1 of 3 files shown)


ButiMM:
## CI verified on the exact head `79318bf`

Run [33054425833](https://github.com/ButiMM/KamLife-Coach/actions/runs/33054425833/job/98457545670), completed 08:33 UTC.

```
suites: 32/35 green, 1 covered nothing in 87s
✓ tracking-contract-tests
✓ production-parity
```

**Baseline-equivalent.** Every number identical to main `5e38582`. The two RED suites are the pre-existing governors — no budget raised, no suppression, no new module.

Ready for the four questions.

---

PR #85, opened 2026-08-27
The distance answer contains the distance

ButiMM:
## The customer failure

Traced through `handleMessage` on main `31444ce`, before any change. A client holding 92kg with an 85kg target:

```
Kam — Scale: down 6.0kg since you started. This week: 0/3 sessions.
Today's protein: 200g / 180g.

That's the distance. The next move is still one thing.
```

The target is never named. The gap is never named. The reply asserts *"that's the distance"* and contains none.

## One correction to the framing

The master map has this as *"the existing trajectory capability is bypassed by the miscellaneous goal-distance mouth."* The trace says otherwise, and the difference matters for the fix.

There is a **dedicated `GOAL_DISTANCE` handler** (`server/handlers/misc-commands.ts:81`) with its own regex and its own `logChat(…, "GOAL_DISTANCE")`, carrying this comment:

> *Distance to goal is a progress-truth question, not the daily-direction card. "How far am I from my goal???" must not fall through to GPT or to the plan menu.*

It claims the question **correctly**. This is not a claim-precedence defect like #81 — the right owner answered, and recited adjacent progress instead of the fact it was asked for. So the fix is not routing; it is giving the owner the fact.

## The correction

`getWeightTruth` (`day-ledger.ts:450`) is the documented single reader of what the scale says — *"ONE DEFINITION OF WHAT THE SCALE SAYS"* — and already held both halves: the latest weigh-in, and the client's target on the user row. It now returns `toGoalKg`, with the same sign convention as the `changeKg` beside it (negative means down the scale). `GOAL_DISTANCE` states it, and the existing recital follows:

```
Kam — 7.0kg to go: 92kg now, 85kg the goal. Scale: down 6.0kg since you started.
This week: 0/3 sessions. Today's protein: 200g / 180g.

That's the distance. The next move is still one thing.
```

**Computed at the owner, not the mouth.** Three inline copies of this arithmetic already exist — `trajectory.ts:108`, `handlers/weight.ts:433`, `scheduler/jobs/monday.ts:107` — and a fourth at a call site is how there comes to be a fifth.

No new handler, no new module. The governor sits at `modules 239/239`, where a new module would be a **new** failure rather than a pre-existing one. `+39 / −1` across two files.

## Controls

| control | result |
|---|---|
| delete the mouth's distance line | **RED** — reproduces the trace verbatim |
| make the owner return `toGoalKg: null` | **RED** — the owner half is load-bearing |
| compute a distance with no target set | **RED** — `"12.0kg to go: 92kg now, 0kg the goal"` |
| withheld scale (`do_not_mention`) | **over-determined — see below** |

### The one control that does not prove what it looks like it proves

The withheld-scale assertion checks that a client who asked us to drop the scale is never told a distance. It passes — but it does **not** prove the guards in this diff are load-bearing. I removed the mouth's `!scaleOff` check *and* made the owner return a distance despite `withheld: true`, and the suite stayed green.

The reason: `"kg"` is a member of the weight topic cluster (`reply-verifier.ts:837`), so the outbound forbidden-topic stripper removes any sentence containing it for that client. The withheld case is protected by **three** independent layers and the third alone suffices.

The assertion is kept because the customer-visible outcome is worth pinning. It is labelled here as an outcome assertion, not a guard proof, rather than being presented as a fourth biting control.

## A false positive worth recording

My first trace showed `"Scale: up 6.0kg since you started"` for a client who went 98 → 92kg — an apparent sign inversion. It was my fixture: I seeded weigh-ins newest-first, but `getWeightTruth` queries `.orderBy(weightLogs.loggedAt)` — ascending — and the stub does not sort. The product is correct. The suite's fixture carries a comment so the next person does not re-seed it backwards.

## Suite state

```
✓ tracking-contract-tests
suites: 32/35 green, 1 covered nothing in 100s
```

Governors, unchanged from main `31444ce`:

```
```

No budget raised, no suppression, no new module.

## Found while tracing — NOT fixed here

Two other phrasings of the same customer intent, both reproduced, both outside this cut:

- `"how much more do I need to lose"` is claimed by the calorie/macro card owner and answered with `2000/2700 kcal` and a food card — a food answer to a weight question. This *is* a claim-precedence defect, the #81 shape.
- `"am I on track for my goal?"` reaches the trajectory door at `routes.ts:1328`, which returns null offline because `getTrajectoryForUser` runs raw `pool.query`. **Unverifiable offline** — not claimed as a defect either way.

Neither is touched here.

---

### DIFF
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -522,6 +522,76 @@ const MUST_WRITE: [string, string][] = [
   }
 
   /**
+   * THE ANSWER MUST CONTAIN THE FACT IT WAS ASKED FOR (2026-08-27, live phone trace).
+   *
+   * "How far am I from my goal?" is claimed correctly — GOAL_DISTANCE in misc-commands owns it and
+   * says so in its own comment: "must not fall through to GPT or to the plan menu". It then answers
+   * a different question. Measured on main 31444ce, for a client holding 92kg with an 85kg target:
+   *
+   *     Kam — Scale: down 6.0kg since you started. This week: 0/3 sessions.
+   *     Today's protein: 200g / 180g.
+   *
+   *     That's the distance. The next move is still one thing.
+   *
+   * The target is never named. The gap is never named. The reply asserts "that's the distance" while
+   * containing no distance. This is not a claim-precedence defect like the meal request above — the
+   * right owner answered — it is the owner reciting adjacent progress instead of the fact asked for.
+   *
+   * The gap now comes from getWeightTruth, the documented single reader of what the scale says,
+   * which already held the current weight and the client's target. Graded on the reply the client
+   * reads, with both directions of over-fire controlled: no target and a withheld scale must each
+   * leave the answer exactly as it was.
+   */
+  {
+    const goalTurn = async (message: string, opts?: { target?: string | null; doNotMention?: string }) => {
+      freshTurn();
+      // Ascending by loggedAt — getWeightTruth orders that way and the stub does not sort, so a
+      // reversed seed reads as a 6kg GAIN and would have this suite grading a fiction.
+      const weighIns = Array.from({ length: 9 }, (_, i) => ({
+        id: `w${i}`, userId: USER.id, weight: 98 - i * 0.75, weightKg: 98 - i * 0.75,
+        loggedAt: new Date(dayStart.getTime() - (8 - i) * 7 * 86_400_000),
+      }));
+      g.__KAMLIFE_STUB_USER = {
+        ...USER, todayWater: "0", currentWeight: 92,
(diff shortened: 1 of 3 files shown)


ButiMM:
## CI verified on the exact head `7eebd61`

Run [33055986941](https://github.com/ButiMM/KamLife-Coach/actions/runs/33055986941/job/98462778816), completed 08:55 UTC.

```
suites: 32/35 green, 1 covered nothing in 108s
✓ tracking-contract-tests
✓ production-parity
```

**Baseline-equivalent.** Every number identical to main `31444ce`. The two RED suites are the pre-existing governors — no budget raised, no suppression, no new module. `ownership: OK` holds with the new field on `getWeightTruth`.

Ready for the four questions. The place to push hardest is the withheld-scale control, which is documented above as an outcome assertion rather than a biting control.

---

PR #86, opened 2026-08-27
Two ways of asking the same plate question reach one owner

ButiMM:
## The customer failure

Traced through `handleMessage` on main `6d6b92f`. Two phrasings of one question, two mouths:

| message | answer |
|---|---|
| `"what can I eat?"` | 🍽️ Next Meal Suggestion, against today's remaining calories |
| `"what should I eat?"` | *Your 3-Day Meal Plan* — with a weekly grocery budget |

## The mechanism

The same shape as the meal-for-calories defect (#81): a door **~9 handlers earlier** in the pipeline claimed it first.

`early-commands.ts:956` builds `isMealPlanRequest` from an exact-match list and a regex. Every term in both names a plan — *meal plan, eating plan, diet plan, weekly meals, nutrition plan* — except one, which names no plan at all: `"what should i eat"`. Meanwhile the door that owns the plate-ask listed only `what should i eat next`, so the bare form matched nothing there and fell through to the plan door.

## The correction

**Two subtractions. No new vocabulary.**

1. The plan door drops `"what should i eat"` from both its exact-match list and its regex, standing down for the bare plate-ask while keeping every phrasing that actually requests a plan.
2. The owner's recogniser drops a `next` suffix it never needed — `what should i eat next` → `what should i eat` — so the bare and suffixed forms are one case.

This serves the invariant directly rather than accreting synonyms: **same customer meaning → same claimant → same decision owner.**

## After

```
"what can I eat?"         -> 🍽️ Next Meal Suggestion — "You need 100g more protein today"
"what should I eat?"      -> identical
"what should i eat"       -> identical
"what should I eat next?" -> identical
"I'm hungry"              -> identical

"give me a meal plan"     -> *Your 3-Day Meal Plan*
"send me a meal plan"     -> *Your 3-Day Meal Plan*
"my meal plan"            -> *Your 3-Day Meal Plan*
"eating plan"             -> *Your 3-Day Meal Plan*
```

They agree on the **content**, not merely the claimant — the converged answer is computed against the client's real remaining protein.

## Controls — four, each failing independently

| control | result |
|---|---|
| restore the term to the plan **regex** | **RED** — `"what should I eat?"` splits back to the meal plan |
| restore it to the **exact-match list** | **RED** — the *unpunctuated* form splits back |
| restore the `next` suffix on the owner | **RED** — the bare form leaves the owner |
| gut the plan vocabulary instead | **RED** — `"eating plan"` loses its answer entirely |

[...]

PR #87, opened 2026-08-27
Coach Health: the adjudicated failures, counted from turns we already record

ButiMM:
## The problem

Every defect this product has fixed arrived the same way: the founder saw it on his own phone, screenshotted it, and explained it. That makes one person's attention the limit of what the product can discover about itself.

`turn_ledger` has recorded the mechanism of every turn since 2026-08-10, and `admin-turns.ts` could already read it — but only a **human verdict** (`failure_category`) put a turn in the queue. Nothing was ever found that nobody looked at.

This adds the missing half: **rule-based detection over the turns already stored.**

## What it does

```
GET /api/admin/coach-health?days=N   ->  clusters, counts, clients, examples
client/src/pages/dashboard.tsx       ->  one card, expandable into the turns
```

```
🩺 Coach Health
1,842 turns · 23 flagged · 3 patterns recurring        [Today] [7d] [30d]

[Claim]    Plate-ask reached the meal-plan owner        #86   11  of 47 asked · 8 clients
[Response] Distance question answered without distance  #85    7  of 12 asked · 5 clients
[Coaching] Step raise ended in a receipt, no next move  #84    5  of 88 asked · 4 clients
[Claim]    Meal request answered with a calorie readout #81    0  of 31 asked
```

Expanding a row shows the offending turns — timestamp, build SHA, the client's words, the reply — each linking into the existing `/admin/turns` triage page.

## Four rules, all already adjudicated

No model judges anything. Each rule is the **same property its cut's contract test asserts**, pointed at production instead of a fixture — from #81, #84, #85, #86.

That inverts what a non-zero count means: it is not a new discovery, it is **a merged fix that is not holding in production.**

A rule is added only *after* adjudication — never when a failure is suspected.

## Honest by construction, in two places

**ASKED is shown beside FAILED.** A rule at `0 of 0 asked` was not exercised in the window. That is *untested*, not healthy, and the page says so. This is where a dashboard starts lying.

**#83 is excluded, and the page says why.** The closed-day card contradiction is a line of text rendered into a PNG; the ledger stores the marker, not the pixels. A rule matching the marker would count *cards*, not contradictions — a number that looks like evidence and isn't. It stays a contract-suite property until the card's next-move line is recorded on the turn.

## Proof

Rules are graded against the **strings actually observed in today's traces**: the pre-fix reply must flag, the post-fix reply must not.

| control | result |
|---|---|
| a rule's `asks` never matches | **RED** — would have reported permanent zeros |
| a rule's `failed` always false | **RED** — "healthy" forever |
| a rule deleted from the library | **RED** — names the lost rule |
| a rule flags everything | **RED** — permanent false positive |

The first two are the dangerous direction: **a blind rule fails by looking good.** Both now break the suite.

## Architecture

**No new module.** The governor counts files under `server/` and `shared/` and sits at exactly `239/239`, so a new file is a *new* failure, not a pre-existing one. The rules live in `admin-turns.ts` — already the reader of `turn_ledger` for failure analysis — beside the queue they feed. The UI is under `client/`, which the governor does not count. File is 526/1200 lines.

**No new telemetry, no second dashboard, no new taxonomy.** `failure_category` and `lifecycle_status` remain the human verdict on top.

```
suites: 32/35 green, 1 covered nothing in 144s
```

Baseline-equivalent — identical to main `88bd787`.

## Stated limits

**This V1 detects only failures already fixed.** It answers *"are our fixes holding?"*, not *"what else is broken?"*. It will not discover a novel failure class, so the founder remains the sensor for unknown failures until enough turns accumulate for clustering to be worth building.

**The rules read `turn_ledger` only.** `chat_history.intent` holds the claimant stamp and would name the wrong owner directly, but the two tables have no foreign key — they join on `user_id` plus timestamp proximity. V1 expresses every rule over `input_text` / `reply` / `mutations`, which one row already carries, rather than resting counts on a fuzzy join.

**Coupling accepted deliberately:** a rule library naturally wants its own module. That would cost a governor raise, which the repo's own ledger shows requires a written account that no existing owner could hold it — and `admin-turns.ts` is exactly such an owner.

---

### DIFF
--- a/client/src/pages/dashboard.tsx
+++ b/client/src/pages/dashboard.tsx
@@ -378,10 +378,14 @@ interface CoachHealthPayload {
   turns: number;
   flagged: number;
   unresolved: number;
+  historical: number;
+  attribution: string;
   clusters: Array<{
     id: string; label: string; layer: string; fixRef: string; expected: string;
-    occurrences: number; clients: number; asked: number;
+    fixedAt: string; trigger: "request" | "mutation";
+    occurrences: number; historical: number; clients: number; candidates: number;
     examples: Array<{ turnId: string; at: string; version: string | null; input: string; reply: string; status: string | null }>;
+    historicalExamples: Array<{ turnId: string; at: string; version: string | null; input: string; reply: string; status: string | null }>;
   }>;
   cannotSurface: Array<{ id: string; fixRef: string; why: string }>;
 }
@@ -415,8 +419,8 @@ function CoachHealth() {
         <div>
           <h3 className="text-xl font-bold font-display">🩺 Coach Health</h3>
           <p className="text-sm text-muted-foreground mt-1">
-            {data.turns.toLocaleString()} turns · {data.flagged} flagged · {data.unresolved} pattern
-            {data.unresolved === 1 ? "" : "s"} recurring
+            {data.turns.toLocaleString()} turns · {data.flagged} regression
+            {data.flagged === 1 ? "" : "s"} since fix · {data.historical} historical
           </p>
         </div>
         <div className="flex gap-1">
@@ -435,8 +439,9 @@ function CoachHealth() {
 
       {data.flagged === 0 && (
         <div className="text-sm text-muted-foreground mb-4 p-3 rounded-md bg-muted/40">
-          No adjudicated failure recurred in this window. Rules showing <strong>0 asked</strong> below
-          were not exercised at all — that is untested, not proven healthy.
+          No adjudicated failure has recurred since its fix merged. A rule whose denominator is
+          <strong> 0</strong> below was not exercised at all in this window — that is untested,
+          not proven healthy.
         </div>
       )}
 
@@ -459,7 +464,11 @@ function CoachHealth() {
                   {c.occurrences}
                 </span>
                 <span className="text-xs text-muted-foreground">
-                  of {c.asked} asked{c.clients > 0 ? ` · ${c.clients} client${c.clients === 1 ? "" : "s"}` : ""}
+                  {c.trigger === "request"
+                    ? `of ${c.candidates} matching request${c.candidates === 1 ? "" : "s"}`
+                    : `${c.candidates} turn${c.candidates === 1 ? "" : "s"} scanned`}
+                  {c.clients > 0 ? ` · ${c.clients} client${c.clients === 1 ? "" : "s"}` : ""}
+                  {c.historical > 0 ? ` · ${c.historical} before the fix` : ""}
                 </span>
               </div>
             </button>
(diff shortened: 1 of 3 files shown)


ButiMM:
## CI verified on the exact head `97d3d1c`

Run [33071358729](https://github.com/ButiMM/KamLife-Coach/actions/runs/33071358729/job/98514160744), completed 12:22 UTC. Run number 550, attempt 1.

**Head SHA confirmed from the run record**, not inferred from timing:

```
head_branch  coach-health
head_sha     97d3d1c53e37ed700e1e5044d1141b97557268bd
event        pull_request
```

```
suites: 32/35 green, 1 covered nothing in 113s
✓ tracking-contract-tests
✓ production-parity
```

**Baseline-equivalent.** Every number identical to main `88bd787`.

The counters worth reading for this cut specifically: `modules` did not move (no new file under `server/`), and `authorshipPoints` held at **425** despite a new endpoint and a new dashboard section. `ownership: OK` — the rules did not create a second reader of any declared fact.

Ready for the four questions. The two worth pushing hardest on:

- **Can the dashboard claim something it isn't measuring?** The blind-rule controls are the answer, because a rule that never matches fails by displaying zeros — it looks like good news. Both directions break the suite, graded against strings observed in the traces rather than paraphrases of them.
- **Operator value.** V1 only detects failures already fixed. It answers "are our fixes holding", not "what else is broken", and that limit is on the PR and on the page.

---

ButiMM:
## Two truthfulness corrections applied — CI verified on `95b695a`

Run [33088925151](https://github.com/ButiMM/KamLife-Coach/actions/runs/33088925151/job/98576122183), completed 15:40 UTC. Run 551, attempt 1.

**Head SHA confirmed from the run record:**

```
head_branch  coach-health
head_sha     95b695a8858b767b711971d351def8470441a327
event        pull_request
```

```
suites: 32/35 green, 1 covered nothing in 83s
✓ tracking-contract-tests   ✓ production-parity
```

**Baseline-equivalent.** Identical to main `88bd787`, across both commits.

---

### 1. Temporal attribution

Each rule now carries `fixedAt` — the **commit date of its squash merge**, read from git, not estimated:

| rule | fix | merged |
|---|---|---|
| meal-for-calories-claim | #81 | `2026-08-27T09:19:40+02:00` |
| step-raise-no-move | #84 | `2026-08-27T10:37:40+02:00` |
| goal-distance-missing | #85 | `2026-08-27T11:03:23+02:00` |
| plate-ask-routing | #86 | `2026-08-27T12:14:47+02:00` |

Hits split into `occurrences` (at or after the merge) and `historical` (before it). **The two are never summed** — the header reads `23 regressions since fix · 41 historical`, never `64 flagged`.

**`version` is not sufficient for this, and the payload says so rather than burying it.** `turn_ledger` stores the build SHA that served each turn, but nothing at runtime knows which SHAs *contain* a given fix — that is git ancestry, and there is no deployments table. Attribution is therefore **merge-time**, declared in the response as `attribution: "merge-time"`, with the deployment-lag caveat carried alongside: a turn shortly after a merge may still have run the old build, so read the build SHA on the example before believing a fresh regression. The build SHA is already on every example row. Deploy-relative attribution is a separate cut, not a caveat to bury here.

### 2. Denominator language

Rules declare `trigger: "request" | "mutation"`, and the label follows from it:

```
[Response] Distance question answered without distance   7  of 12 matching requests · 5 clients
[Coaching] Step raise ended in a receipt, no next move   5  88 turns scanned · 4 clients
```

No scanned turn is called "asked" any more.

### Controls — four new, each failing independently

| control | result |
|---|---|
| remove the temporal split | **RED** — "would report a fix as not holding using the evidence that justified it" |
| treat everything as historical | **RED** — "a genuine recurrence would be filed as history and never surface" |
| unparseable `fixedAt` | **RED** — names the rule |
| missing `trigger` | **RED** — "its denominator cannot be labelled honestly" |

The second is the trap in fixing the first: over-correcting hides real regressions and **fails looking clean**. Both sides are graded one second either side of each merge instant, so the boundary itself is under test rather than a comfortable distance from it.

`isRegression` is exported for exactly this reason — the property the page rests on has to be gradeable, not buried in a route handler.

### Unchanged

Existing dashboard · existing `turn_ledger` · no GPT judging · no new telemetry · no second dashboard · no new taxonomy · drill-down to `turnId` · build SHA shown · #83 still marked unmeasurable from ledger data · `modules` unmoved.

### Commits

| SHA | scope |
|---|---|
| `97d3d1c` | Coach Health V1 — rules, endpoint, dashboard card |
| `95b695a` | temporal attribution + denominator language |

Ready for the four-question attack again.

---

PR #89, opened 2026-08-28
Coach Health V2: the failures nobody has adjudicated yet

ButiMM:
V1 watched four failures we had already fixed. That answers *"are our fixes holding"* and nothing about **what else is broken** — so the founder was still the only sensor for anything new. This is the half that finds the unknown.

## Five invariants, over fields the ledger already stores

| invariant | layer | what it watches |
|---|---|---|
| `question-mutated-state` | State | **Law 2** in production — a question must never write |
| `durable-write-no-move` | Coaching | **Law 4**, generalised past steps to every domain that writes |
| `reply-contradicts-itself` | Response | closes the day and sells food in one breath |
| `unowned-message` | Claim | the fallback shape: no door claimed this |
| `empty-reply` | Response | the client received nothing |

## Candidates are evidence, not verdicts

An invariant never says *"this is a defect."* It says a property did not hold on this turn, and attaches the turn. A detector that declared defects would fill the queue with confident nonsense, and every hour spent disproving it is an hour not spent fixing the coach.

## The packet, not a chart

```
CH-3F2A   high   The coach did not understand a message      "hungry what can eat"
          9 clients · 17 turns · candidate

          Layer: Claim · Expected: an owner answers, or the coach asks one specific question back
          First seen 28/08 07:14 · last seen 29/08 19:02

          Client: "I'm at Nando's, what should I order?"
          Coach:  "I didn't catch that one — what was it, roughly?"
          build 70c7d2a · Open the turn →
```

Ranked by **clients first, then recurrence**: nine clients hitting something once is the product misbehaving; one client hitting it nine times is one conversation.

## No second lifecycle

`turn_ledger` already carries `lifecycle_status`, `failure_category` and `fix_ref`, editable through the verdict endpoint this file has had since 2026-08-25. A candidate's status is the **aggregate of its turns**:

```
untriaged        -> candidate
failure_category -> adjudicated
fix_ref          -> engineering
deployed         -> deployed
revalidated      -> resolved
```

Triaging a turn moves the candidate. A parallel candidate store would have been a second source of truth guaranteed to drift from the first.

## No new module, no new table, no model

The governor counts files under `server/` and sits at `239/239`. Detection runs **on read**, so the evidence persists because `turn_ledger` persists — which is what lets a weekend of testing accumulate while every agent on this project is switched off. Nothing here writes, routes, or changes production behaviour.

## The governor caught a real defect before review did

`messageDeciders` went **32 → 33** on the first commit. Two things underneath it, one of them mine:

**The real one.** `question-mutated-state` carried its *own regex* for "is this a question" — a second answer to what `isAskingNotReporting` already owns and the whole tracking contract is built on. A detector for claim duplication, committing claim duplication. It now calls the existing floor.

> **Control:** delete the floor's body and the **contract suite breaks on the product side** — `"Is 2 litres of water enough?"` writes water again. Detector and product fail together because they read one owner.

**The false positive.** The guard greps the literal `.test(m)`, `m` being the conventional name for the client's message. Two mutation loops read `.some(m => DURABLE_MUTATION.test(m))` — testing a mutation string, not a message. Renamed to `mut`; behaviour unchanged. Back to **32/31 by losing a false positive rather than moving a ceiling** — the precedent the ledger set when `authorshipPoints` returned to 420 the same way.

## Controls — seven, each failing independently

| control | result |
|---|---|
| invariants always hold | **RED** — "this turn would never reach the queue" |
| an invariant never holds | **RED** — "the queue fills with noise and stops being read" |
| drop an invariant | **RED** — names the lost property |
| `question-mutated-state` goes blind | **RED** — names the missed turn |
| delete the shared asking floor | **RED** — *the product* breaks, not just the detector |
| collapse the cluster signature | **RED** — "would send an engineer after a pattern that does not exist" |
| unstable `CH-` handle | **RED** — "would not survive being written down" |

The first two are the two ways a detector lies, **and only one of them looks like failure.** Clustering is graded both directions: punctuation variants of one question must merge, two unrelated questions must not.

## Suite

```
suites: 32/35 green, 1 covered nothing in 101s
```

Baseline-equivalent — identical to main `70c7d2a`.

## Stated limit

`turn_ledger` **does not store the claimant.** So "the wrong door answered" is visible only where the reply betrays it — a plan when a plate was asked for, a fallback where an owner exists. A claim defect whose reply looks plausible is invisible here. Stamping the claimant on the turn is the cut that fixes that, and it is not this one.

Nothing here claims to wake an agent up. The queue accumulates; invoking the engineer remains a human action.

---


ButiMM:
## Corrected tree `a3519ed` — CI verified, every governor compared

Run [33149514572](https://github.com/ButiMM/KamLife-Coach/actions/runs/33149514572/job/98777989133), run 554, attempt 1, completed 06:56 UTC.

```
head_branch  coach-health-v2
head_sha     a3519ed628740a611fe3465e00109f201c213515
event        pull_request
```

### First: my previous report on this PR was wrong

I called `1b628cb` baseline-equivalent. It was not — CI carried a **fourth** red counter that main does not have:

```
✗ regexLiterals: 452 — budget 449
```

The miss matters more than the number. I had grepped my local run for `messageDeciders|looksLikePredicates|authorshipPoints|ownership`, so a counter I had not thought to name **could not appear in my output**, and I read a filtered view as the whole result.

### Every governor, diffed — not a chosen subset

Both guards run on `70c7d2a` and on `a3519ed`, full output compared:

```
ARCHITECTURE GUARD: full diff main -> branch
IDENTICAL — no counter moved, named or unnamed

FILE-SIZE GUARD: full diff main -> branch
IDENTICAL
```

Byte-identical, so a counter I never thought to check cannot hide in it.

```
suites: 32/35 green, 1 covered nothing in 91s
✓ tracking-contract-tests   ✓ production-parity
```

### What changed to get there

**`DURABLE_MUTATION` was a second answer.** "Was this a durable write?" is owned by `durableDomains` — the same reader `closeCoachingTurn` consults before deciding a turn owes a coaching move. A private regex meant detector and product could disagree about what a write *is*: the exact defect class this detector exists to find.

> **Control:** empty `durableDomains` and **two invariants go blind at once**. Detector and coaching turn fail together, because they read one thing.

**The contradiction invariant was reasoned, not observed.** The real #83 defect is between text and card, and the card is a PNG the ledger cannot see — already documented as unmeasurable. The text-only variant was built from argument and never seen in a trace. Removed with its grading, rather than paid for with a budget.

Four invariants remain, each grounded in a law or an observed trace: **Law 2**, **Law 4**, the fallback reply seen in the hunger trace, and an empty reply.

### For the review — where I would attack this hardest

The lens that just caught a real defect is *does the detector hold the same authority as the product for every fact it claims?* Two answers were no, and both now read the product's owner (`isAskingNotReporting`, `durableDomains`), proven by controls that break product and detector together.

**One invariant still has no owner to defer to: `unowned-message`.** It matches on the fallback reply's *wording*. If that copy is reworded, the invariant silently stops firing — a false negative that presents as health. There is no exported "this reply is the fallback" predicate to reuse. That is the weakest authority claim in this PR and worth attacking first.

Also worth attacking: whether a candidate can be created without a ledger-grounded reason (every candidate requires a failing invariant on a real row, but the clustering signature is deliberately crude), and tester concentration — ranking is by distinct clients, so a single tester cannot manufacture a high-priority candidate alone, but two testers can.

---

ButiMM:
## Four proof obligations — one FAILED and is fixed. Review tree is now `8a7832b`

`a3519ed` is superseded. Obligation 2 failed on it, so that tree should not be reviewed.

### 1. Exact-SHA CI, complete governor output ✅

Run [33150223196](https://github.com/ButiMM/KamLife-Coach/actions/runs/33150223196/job/98780226662), run 555, attempt 1. **Head SHA from the run record:**

```
head_branch  coach-health-v2
head_sha     8a7832bafcc1f2b843a49192a724ee0a8638af50
```

```
suites: 32/35 green, 1 covered nothing in 106s
✓ tracking-contract-tests   ✓ production-parity
```

Both guards run on `70c7d2a` and on `8a7832b`, **full output diffed** — not a chosen subset:

```
ARCHITECTURE: IDENTICAL (every counter, named or not)
FILE SIZES:   IDENTICAL
```

### 2. A seeded genuine failure produces a packet with exact turn ids — ❌ FAILED, now fixed

Driving the **real endpoint** against seeded `turn_ledger` rows — which no unit test had done — exposed a live defect in `a3519ed`. Three clients, one question, the coach understanding none of them:

```
"I'm at Nando's, what should I order?"   ->  "Sorry Kam, I didn't quite catch that 🙂"
"at nandos what should i order"          ->  "Sorry Kam, I didn't quite catch that 🙂"
"Nando's — what should I order?"         ->  "Eish Coach K had a moment. Try that again."
```

produced **three candidates of one client each**:

```
CH-2E29  low  "at nando what should order"   1 client
CH-....  low  "at nandos what should order"  1 client
CH-....  low  "nando what should order"      1 client
```

Split by a possessive and a missing preposition. Ranking is by distinct clients, so all three sat at `low` and **none would ever have surfaced** — the detector missing a real repeated failure by *fragmenting* it. A fragmented queue is indistinguishable from a healthy one, which is the dangerous direction.

Fixed — signature now strips a trailing `s`, keeps the four longest content words, and sorts them. Same three messages:

```
CH-F6B2  medium  "nando order should what"   3 clients · 3 turns · status=candidate
turn ids: t-aaa, t-bbb, t-ccc
client: "I'm at Nando's, what should I order?"
coach:  "Sorry Kam, I didn't quite catch that 🙂"
```

The `durableDomains`-backed invariant fired independently on the same run: `CH-D4BA`, ids `t-ddd`, from a seeded raise with no next move.

### 3. The false case must not become a candidate ✅ (passed first run, unchanged)

Two turns of `"what should I eat this week"`, correctly answered with the week's shopping list:

```
✓ produced NO candidate — no invariant fired, nothing entered the queue
candidates total: 2 — unowned-message, durable-write-no-move
```

### 4. The durable-write signal is literally the product's ✅

```
admin-turns.ts:14   import { durableDomains } from "../understanding/messy-intake"
admin-turns.ts:333  isAskingNotReporting(t.input) && durableDomains(t.mutations).length > 0
admin-turns.ts:341  durableDomains(t.mutations).length > 0 && !lastBlockIsAMove(t.reply)
DURABLE_MUTATION    (none — the copy is gone)
```

> **Control:** empty `durableDomains` in `messy-intake.ts` and **both** invariants go blind at once. Detector and product fail together because they read one thing. If the product says `durableDomains = []`, the detector has no second opinion to fall back on.

### Clustering, both directions controlled

| control | result |
|---|---|
| revert to first-five-words-in-order | **RED** — one question becomes three candidates |
| collapse the signature | **RED** — two unrelated questions become one |

Merging things that are *not* the same question is the worse error, so it is graded in the same block.

### Still the weakest claim, unchanged

`unowned-message` has no product owner to defer to — it matches the fallback reply's *wording*. Reword that copy and the invariant silently stops firing: a false negative presenting as health. No exported "this reply is the fallback" predicate exists to reuse.

---

PR #90, opened 2026-08-28
Proactive coaching gets one answer to "may we prescribe?"

ButiMM:
## The customer failure

Traced through both policy paths on **one** client state — sick, two of seven days logged, 09:00:

```
decideProactive  ->  rest    "Rest today. No training, no targets."
underPolicy      ->  hold    "Nothing new today. Do exactly what you did yesterday."
```

Telling a sick person to repeat yesterday is the wrong instruction, and it re-broke a finding `one-action.ts` already carried: **illness is directly observed durable state, sufficient by construction** (2026-08-18). `decideProactive` knew that through `evidenceFor`; the gate took a bare boolean from its caller and structurally could not.

## Three disagreements found, two closed

| # | disagreement | status |
|---|---|---|
| 1 | `rest` self-evidencing proactively, not reactively | **closed** — one definition, both paths |
| 2 | food floor `>= 3` reactive vs `PROACTIVE_LOG_FLOOR` (4) proactive | **closed** — floor moved to the policy owner |
| 3 | weight evidence counted proactively, ignored reactively | **found, not fixed** — see below |

## The correction

The gate stopped taking a **verdict** and now takes the **evidence**, reaching its answer through the same two functions `decideProactive` uses:

```ts
evidenceFromKind(kind)                                  // rest -> sufficient; investigative -> insufficient
  ?? evidenceFromLedger(foodSufficient, weightSufficient)
```

No third policy, no new authority. `PROACTIVE_LOG_FLOOR` moved to `one-action.ts` — the policy that applies it — and the scheduler imports it rather than defining a second copy. The compiler enumerated all four call sites when the signature changed.

## The four proof obligations

| # | obligation | result |
|---|---|---|
| 1 | sick + thin evidence | ✅ both paths → `rest`; neither says "do yesterday" |
| 2 | sparse but healthy | ✅ no prescription; a fresh event still answered, same on both |
| 3 | sufficient evidence | ✅ both → `protein`; the coach is not mute |
| 4 | illness constrains the **ladder** | ✅ ladder refuses; gate passes the ladder's object through |

**Obligation 4 is the one the cut rests on.** If `chooseAction` could still pick `train` for a sick client and a wrapper rewrote the sentence afterwards, we would have replaced one contradiction with a politer one — the decision still wrong, only the words agreeing. Illness is **rung 2 of the ladder**, ahead of every prescriptive rung.

Graded from three states that **do** reach a prescription when healthy — would-train, would-walk, would-protein — and the fixtures are themselves asserted to still do so, otherwise the test would pass because nothing was going to be prescribed anyway. The gate check is **object identity**, not string equality: a gate that rebuilt an equivalent action would be a second author of the same decision.

## Controls — six, each failing independently

| control | result |
|---|---|
| gate takes a bare verdict again | **RED** — reproduces the trace verbatim |
| `rest` stops being self-evidencing | **RED** — the sick client loses rest |
| gate opens for everyone | **RED** — sparse evidence stops protecting |
| gate holds everyone | **RED** — the coach goes mute |
| illness stops constraining the ladder | **RED** — the paths disagree again |
| gate returns `{ ...action }` | **RED** — re-authoring, not applying |

Both failure directions are covered: a gate that agrees by holding everything is as wrong as one that agrees by prescribing everything.

Law 26 is untouched — insufficient evidence still yields the downgrade to a real question or an honest hold, never an invented explanation.

## Found, not fixed — recorded at the gate, not on a board

**One evidence → one policy** is what this cut establishes. **One policy → one action** is *not*, and on the insufficient branch the paths still differ:

```
sparse but healthy   decideProactive -> weigh  "Stand on a scale this morning."
                     underPolicy     -> hold   "Nothing new today."
```

There is a plausible product reason — proactively the ask *is* the whole message, so silence wastes the contact; reactively the client has already been answered — **but that reason is unproven.** No client has been traced to establish it. It is a product-policy question to settle from production evidence before changing, and the note at the gate names what closing it would cost.

The weight-evidence gap (#3) is recorded the same way, at the reactive call site: the proactive side reads a stall verdict this path never computes, so passing anything but `false` would be inventing evidence.

## A gap-test that broke on wording, not behaviour

The full suite caught `gap-tests` going RED — 31/35 against main's 32/35 — where the guards were byte-identical and would have passed it through. The assertion greps `morning.ts` for the literal `evidenced: false`, which the signature change renamed. The property it guards is intact and now stated twice over.

**Control:** make the degraded fallback claim `foodSufficient: true, weightSufficient: true` and the assertion fails. It still guards what it was written to guard.

## Verification

```
suites: 32/35 green, 1 covered nothing — main's exact baseline
contract suite ✓   gap-tests ✓ 335/335

guards, 9d76633 vs 15dd88c:
  ARCHITECTURE: IDENTICAL (every counter, named or not)
  FILE SIZES:   IDENTICAL
```

The guards were re-run on `15dd88c` rather than carried forward from `6cbcbe7`: `authorshipPoints` counts client-facing strings and the parked-item note quotes two coach lines verbatim, so a comment-only commit could in principle have breached a governor.

## Commits

| SHA | scope |
|---|---|
| `e6bbe78` | the policy consolidation |
| `e32bee6` | test-only: gap-test follows the new shape |
| `6cbcbe7` | obligation 4's proof — ladder + object identity |
| `15dd88c` | the parked divergence, recorded at the gate |

---

### DIFF
--- a/script/gap-tests.ts
+++ b/script/gap-tests.ts
@@ -2895,7 +2895,11 @@ test("sweep: the >7-day client's decision is used, not computed and discarded", 
   assert.ok(/return formatOneAction\(underPolicy\(chooseAction\(\{/.test(morning),
     "a drifting client must not get silence because a ledger read timed out — and the fallback "
     + "must reach the decision owner through the policy contract, not around it");
-  assert.ok(/evidenced: false/.test(morning),
+  // The shape changed on 2026-08-28 when the gate stopped taking a pre-computed verdict and
+  // started taking the evidence itself, so that it and decideProactive reach one answer through
+  // one function. The property this line protects is unchanged: the degraded fallback claims NO
+  // evidence, because the state read is what just failed.
+  assert.ok(/foodSufficient: false, weightSufficient: false/.test(morning),
     "…with evidence stated honestly: the state read is what just failed");
 });
 
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -883,6 +883,43 @@ const MUST_WRITE: [string, string][] = [
         failures.push(`A well-evidenced client was held silent by the gate — it now refuses to coach at all`);
       }
     }
+
+    /**
+     * ILLNESS CONSTRAINS THE LADDER, NOT THE PROSE.
+     *
+     * The distinction the whole cut rests on. If chooseAction could still pick `train` for a sick
+     * client and a wrapper rewrote the sentence afterwards, we would have replaced one
+     * contradiction with a politer one — the decision would still be wrong, and only the words
+     * would agree. Illness is rung 2 of the ladder, ahead of every prescriptive rung, so the
+     * refusal happens where the decision is made.
+     *
+     * Graded from states that DO reach a prescription when healthy, so the assertion cannot pass
+     * because nothing was going to be prescribed anyway.
+     */
+    for (const [label, over] of [
+      ["would train", { workout: { sessionsLast7d: 0, sessionsThisWeek: 0 }, evidence: { foodSufficient: true, weightSufficient: true } }],
+      ["would walk", { steps: { avg7d: 500 }, today: { kcal: 1200, protein: 70, steps: 200, logged: true, hour: 14 }, evidence: { foodSufficient: true, weightSufficient: true } }],
+      ["would protein", { today: { kcal: 1200, protein: 20, steps: 2000, logged: true, hour: 9 }, evidence: { foodSufficient: true, weightSufficient: true } }],
+    ] as [string, any][]) {
+      const healthy = chooseAction(dayStateFrom(stateOf(over), profile, { hour: 9 }));
+      const sick = chooseAction(dayStateFrom(stateOf({ ...over, health: { sick: true } }), profile, { hour: 9 }));
+      if (healthy.kind === "hold" || healthy.kind === "rest") {
+        failures.push(`The "${label}" fixture no longer reaches a prescription when healthy (${healthy.kind}) — the illness assertion below would pass for the wrong reason`);
+      }
+      if (sick.kind !== "rest") {
+        failures.push(`The decision ladder prescribed "${sick.todo}" to a sick client — illness is not constraining the decision, only the wording`);
+      }
+    }
+    // AND THE GATE DOES NOT RE-WORD WHAT THE LADDER DECIDED. Object identity, not string equality:
+    // a gate that rebuilt an equivalent action would be a second author of the same decision.
+    {
+      const s = stateOf({ health: { sick: true } });
+      const raw = chooseAction(dayStateFrom(s, profile, { hour: 9 }));
+      const passed = underPolicy(raw, { foodSufficient: false, weightSufficient: false, dreamGoal: null });
+      if (passed !== raw) {
+        failures.push(`The policy gate returned a different object than the ladder chose — it is re-authoring the decision, not applying a policy to it`);
+      }
+    }
   }
 
   /**
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -812,6 +812,80 @@ const MUST_WRITE: [string, string][] = [
   }
 
   /**
+   * ONE CLIENT STATE -> ONE PROACTIVE ANSWER (2026-08-28, traced through both policy paths).
+   *
+   * underPolicy and decideProactive both answer "may we prescribe?" and answered it differently.
+   * Measured on ONE state — sick, two of seven days logged, 09:00:
+   *
+   *     decideProactive  ->  rest    "Rest today. No training, no targets."
+   *     underPolicy      ->  hold    "Nothing new today. Do exactly what you did yesterday."
+   *
+   * Telling a sick person to repeat yesterday is the wrong instruction, and it re-broke a finding
+   * one-action.ts already carried: illness is directly observed durable state, sufficient by
+   * construction (2026-08-18). decideProactive knew it; the gate took a bare boolean and could not.
+   *
+   * Graded on the ACTIONS BOTH PATHS PRODUCE from one state, not on the predicate that chooses
+   * them — the two must agree, and the sparse-evidence protection must survive.
+   */
+  {
+    const { chooseAction, underPolicy, decideProactive, dayStateFrom } = await import("../server/one-action");
+    const profile: any = {
+      dreamGoal: null, biggestStruggle: null, doNotMention: null, lifeContext: null,
+      weeksOnProgramme: 5, sessionsTarget: 3, calorieTarget: 2700, proteinTarget: 180, stepsTarget: 10000,
+    };
+    const stateOf = (over: any) => ({
+      name: "Kam", goalType: "fat_loss",
+      health: { sick: false },
+      food: { loggedDays7d: 2, daysSinceAnyLog: 0 },
+      workout: { sessionsLast7d: 1, sessionsThisWeek: 1 },
+      steps: { avg7d: 3000 },
+      weight: { daysSinceWeighIn: 4, trendUsable: false },
+      today: { kcal: 1200, protein: 70, steps: 2000, logged: true, hour: 9 },
+      evidence: { foodSufficient: false, weightSufficient: false },
+      ...over,
+    }) as any;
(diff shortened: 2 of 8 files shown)


ButiMM:
## Exact-SHA CI verified on `15dd88c`

Run [33175721012](https://github.com/ButiMM/KamLife-Coach/actions/runs/33175721012/job/98863693227), run 557, attempt 1, completed 13:33 UTC.

```
head_branch  policy/one-constitution
head_sha     15dd88c3b25904df48d22a3eb26985353d4c8c66
event        pull_request
```

```
suites: 32/35 green, 1 covered nothing in 111s
✓ tracking-contract-tests   ✓ production-parity   ✓ gap-tests
```

**Baseline-equivalent.** Locally both guards were run on `9d76633` and on `15dd88c` and the **complete output diffed** — not a chosen subset:

```
ARCHITECTURE: IDENTICAL (every counter, named or not)
FILE SIZES:   IDENTICAL
```

`gap-tests` is green here, which is the suite that caught this cut's only real break — a source-string assertion that failed on wording while the guards stayed byte-identical.

### Ready for the four-question attack on this exact tree

The claim under review, stated precisely so it can be attacked:

- **Established:** one evidence → one policy. Both paths reach the verdict through `evidenceFromKind` / `evidenceFromLedger`, and `underPolicy` returns the ladder's own object rather than re-authoring it.
- **Not established, deliberately:** one policy → one action, on the insufficient branch. Recorded at the gate in `one-action.ts`, not on a board.

Where I would attack hardest, in order:

1. **Obligation 4's fixtures.** The illness assertions only mean something if the would-train / would-walk / would-protein states genuinely reach a prescription when healthy. That precondition is itself asserted — but if a fixture drifts to `hold` for an unrelated reason, the guard against it is one line, and worth checking.
2. **`evidenceFromKind` is now the single owner of "is this action self-evidencing", and `rest` is its only `sufficient` entry.** If another kind should be self-evidencing (`come_back` follows from an observed absence, and the file's own comment says so) then the consolidation is incomplete in a direction the controls cannot see.
3. **The reactive `weightSufficient: false`.** Honest, but it means the reactive path is strictly more conservative than the proactive one for any client whose evidence is weight-only.

---

ButiMM:
## The three code sections, verbatim from `15dd88c`

Posted because the reviewer cannot read the private repo — these should be checked as code, not taken on my description.

---

### 1. The illness test, and why it cannot pass vacuously

`script/tracking-contract-tests.ts`:

```ts
for (const [label, over] of [
  ["would train", { workout: { sessionsLast7d: 0, sessionsThisWeek: 0 }, evidence: { foodSufficient: true, weightSufficient: true } }],
  ["would walk", { steps: { avg7d: 500 }, today: { kcal: 1200, protein: 70, steps: 200, logged: true, hour: 14 }, evidence: { foodSufficient: true, weightSufficient: true } }],
  ["would protein", { today: { kcal: 1200, protein: 20, steps: 2000, logged: true, hour: 9 }, evidence: { foodSufficient: true, weightSufficient: true } }],
] as [string, any][]) {
  const healthy = chooseAction(dayStateFrom(stateOf(over), profile, { hour: 9 }));
  const sick    = chooseAction(dayStateFrom(stateOf({ ...over, health: { sick: true } }), profile, { hour: 9 }));
  if (healthy.kind === "hold" || healthy.kind === "rest") {
    failures.push(`The "${label}" fixture no longer reaches a prescription when healthy (${healthy.kind}) — the illness assertion below would pass for the wrong reason`);
  }
  if (sick.kind !== "rest") {
    failures.push(`The decision ladder prescribed "${sick.todo}" to a sick client — illness is not constraining the decision, only the wording`);
  }
}
```

Two things to check here:

- It calls **`chooseAction`** — the ladder — not `underPolicy`. So it tests where the decision is *made*, not where prose is applied.
- The `healthy` assertion is the anti-vacuity guard: if a fixture ever stops reaching a prescription, the suite says so **instead of** letting the `sick` assertion pass for the wrong reason.

**Control, run on this tree** — disable the ladder's illness rung (`if (false && s.sick)`):

```
✗ Two proactive constitutions disagree on one client: decideProactive says
  "Stand on a scale this morning, before you eat." and underPolicy says
  "Nothing new today. Do exactly what you did yesterday."
✗ A sick client did not get rest from the reactive gate — it said
  "Nothing new today. Do exactly what you did yesterday."
```

And the separate identity assertion:

```ts
const raw    = chooseAction(dayStateFrom(s, profile, { hour: 9 }));
const passed = underPolicy(raw, { foodSufficient: false, weightSufficient: false, dreamGoal: null });
if (passed !== raw) {
  failures.push(`The policy gate returned a different object than the ladder chose — it is re-authoring the decision, not applying a policy to it`);
}
```

**Control** — make the gate return `{ ...action }` instead of `action`:

```
✗ The policy gate returned a different object than the ladder chose — it is re-authoring the decision, not applying a policy to it
```

---

### 2. `evidenceFromKind` — only `rest` is self-evidencing

`server/one-action.ts`:

```ts
function evidenceFromKind(kind: ActionKind): DecisionEvidence | null {
  if (kind === "rest") return "sufficient";
  if (INVESTIGATIVE.has(kind)) return "insufficient";
  return null;   // nothing follows from the kind alone — ask the ledger
}

/** The ledger half. Either source on its own is enough; that is one rule, stated once. */
function evidenceFromLedger(foodSufficient: boolean, weightSufficient: boolean): DecisionEvidence {
  return foodSufficient || weightSufficient ? "sufficient" : "insufficient";
}

function evidenceFor(s: ProactiveStateForDecision, kind: ActionKind): DecisionEvidence {
  return evidenceFromKind(kind)
    ?? evidenceFromLedger(s.evidence.foodSufficient, s.evidence.weightSufficient);
}
```

There is **no loose fallback**. The third branch returns `null`, not `"sufficient"` — an unknown kind falls through to the ledger, it does not get waved past.

Every `"sufficient"` literal in the whole file:

```
638:  evidence: "sufficient" | "insufficient";          <- the type
750:  if (!PRESCRIPTIVE.has(action.kind) || verdict === "sufficient") return action;
767:  if (kind === "rest") return "sufficient";          <- the only kind-based grant
774:  return foodSufficient || weightSufficient ? "sufficient" : "insufficient";
```

`rest` at line 767 is the only place a kind is granted sufficiency.

---

### 3. `weightSufficient: false` means "no evidence", and leads only to a hold

Both reactive call sites, identical:

```ts
} as any), { foodSufficient: truth.window.daysLogged >= PROACTIVE_LOG_FLOOR,
     // WEIGHT EVIDENCE IS NOT COUNTED HERE, and that is a known gap rather than a
     // decision: the proactive side reads a stall verdict this path never computes, so
     // passing anything but false would be inventing evidence. It means a client with a
     // usable weight trend and a thin food log is still held on this path.
     weightSufficient: false, dreamGoal: user.dreamGoal });
```

The gate is the only consumer, and `false` can only reduce the verdict:

```ts
export function underPolicy(
  action: OneAction,
  opts: { foodSufficient: boolean; weightSufficient: boolean; dreamGoal?: string | null },
): OneAction {
  const verdict = evidenceFromKind(action.kind)
    ?? evidenceFromLedger(opts.foodSufficient, opts.weightSufficient);
  if (!PRESCRIPTIVE.has(action.kind) || verdict === "sufficient") return action;
  return holdAction(opts.dreamGoal);
}
```

`evidenceFromLedger` is a disjunction, so `weightSufficient: false` can only ever make the verdict **less** permissive — it can never select a different action. The single consequence:

```ts
function holdAction(dream?: string | null): OneAction {
  return {
    kind: "hold",
    todo: "Nothing new today. Do exactly what you did yesterday.",
    why: why("This is the part that works. Boring and repeated beats clever and occasional.", dream),
  };
}
```

So `false` means *"we do not have this evidence"* → at most a hold. It never routes to some other action.

All four `underPolicy` call sites, for completeness:

```
server/handlers/misc-commands.ts:690        reactive
server/understanding/live.ts:92             reactive
server/scheduler/jobs/morning.ts:63         proactive degraded fallback
server/scheduler/proactive-decision.ts:101  proactive degraded fallback
```

The two fallbacks pass `foodSufficient: false, weightSufficient: false` — the state read is what just failed, so neither source is claimed.

---

The unresolved **insufficient-evidence → action** divergence is deliberately parked and not touched by this PR.

---

PR #94, opened 2026-08-28
CTO one-shot: bind coaching outputs to existing truth

ButiMM:
CTO one-shot implementation branch from `a633c7c`.

Includes the repository-corrected implementation discussed with Grok:
- weight trend output is bound to existing `weightTrendUsable()` evidence and durable health state;
- response gate recognizes scale/weight direction claims;
- re-entry voice uses canonical `contactState`;
- workout session reports capture client-stated weekly ordinal + day type and use weekly ledger count;
- protein-gap meal suggestions materially honor large gaps;
- exact ordinal session regression added to existing tracking suite;
- one-shot implementation spec retained in docs.

This is intentionally not a second architecture rewrite. Grok/Claude review should attack the diff and the remaining final-authority convergence before merge.

### DIFF
--- a/script/gap-tests.ts
+++ b/script/gap-tests.ts
@@ -2895,7 +2895,11 @@ test("sweep: the >7-day client's decision is used, not computed and discarded", 
   assert.ok(/return formatOneAction\(underPolicy\(chooseAction\(\{/.test(morning),
     "a drifting client must not get silence because a ledger read timed out — and the fallback "
     + "must reach the decision owner through the policy contract, not around it");
-  assert.ok(/evidenced: false/.test(morning),
+  // The shape changed on 2026-08-28 when the gate stopped taking a pre-computed verdict and
+  // started taking the evidence itself, so that it and decideProactive reach one answer through
+  // one function. The property this line protects is unchanged: the degraded fallback claims NO
+  // evidence, because the state read is what just failed.
+  assert.ok(/foodSufficient: false, weightSufficient: false/.test(morning),
     "…with evidence stated honestly: the state read is what just failed");
 });
 
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -812,6 +812,117 @@ const MUST_WRITE: [string, string][] = [
   }
 
   /**
+   * ONE CLIENT STATE -> ONE PROACTIVE ANSWER (2026-08-28, traced through both policy paths).
+   *
+   * underPolicy and decideProactive both answer "may we prescribe?" and answered it differently.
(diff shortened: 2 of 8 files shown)


ButiMM:
Grok review requested against the actual diff, not the spec narrative. Attack order: (1) any remaining coaching writer can bypass the canonical/evidence path; (2) 16:49 illness-adjacent trend still able to produce direction/"keep fuelling"; (3) session ordinal/time identity and any stale writer; (4) re-entry proxy regression; (5) large protein gap tool contract; (6) any parallel decision authority still reachable. Do not request symptom patches. Treat the PR as one convergence cut and report substantive blockers only.

ButiMM (commented):
Grok review contract for this one-shot implementation:

Attack the whole diff, not just the three original screenshots.

1. Coaching turns must have one final authority. Any handler that can still finish a coaching turn with a user-facing string before the shared boundary is a blocker.
2. Weight trend must use the existing evidence owner. If trend evidence is unusable, no direction or compensating weight advice may survive later in the turn.
3. Re-entry must follow canonical contact state, not logging density.
4. Workout identity must come from the current-turn statement + authoritative ledger; do not accept invented lifetime/session copy.
5. Meal recommendations must honor the stated protein gap/priority; do not accept a new hardcoded menu as a substitute for a tool contract.
6. Look specifically for any remaining second decision world, second LLM writer, withNextMove/closeCoachingTurn tail, card-level coaching act, or GPT fallback that can contradict the chosen action.
7. Reject symptom guards that leave the old writer alive.
8. Reject any coaching TurnOutcome that carries a finished reply string.
9. T1/T2/T3/T6 and the existing contract suite must remain biting. Try to break them by bypassing the obvious path.
10. Do not widen into patent, trademark, POPIA, new memory/event architecture, or generic Pulse redesign. The question is whether this branch makes the existing truth/policy machinery reliably control the customer-facing coach.

Approve only the law, not the narrative.

ButiMM (commented):
CTO implementation checkpoint: triggering the repository-side patch runner for the structured weight/coaching seam. This is preparatory work; authority convergence is not claimed until the coaching early-return is gone and T1/T4/T8 pass.

ButiMM (commented):
Re-run implementation checkpoint against current #94 hunk. The temporary patch runner should now convert the weight/trend claimant to a structured CoachingTurn and route it through the existing live boundary; it must remove its own temporary files.

ButiMM (commented):
Re-triggering repository-side convergence patch against the current #94 head. This runner must convert the weight coaching return to structured CoachingTurn and route it through the existing authority boundary; it must remove its temporary artifacts. No authority-convergence claim until routes early-return is actually closed.

PR #101, opened 2026-08-31
Three failures from the observed trace: the protein gap, the weight contradiction, the unanswerable question

ButiMM:
Three of the five failures from the live trace of **2026-08-28 16:45–16:49 SAST**. Each fixed against evidence the code already held. Defects 2 and 3 are **not traced** and are not touched.

---

## Defect 1 — the recommendation ignored the number it had just quoted

```
Coach: "You need 129g more protein today. That is the priority."
       "1. 2 eggs + pap (~300 kcal, 18g protein)"
```

with **2,146 kcal unspent**. `misc-commands.ts` printed `protLeft` and never read it again, and never read `calLeft` at all — a 21g gap and a 129g gap produced identical text, because `needsProtein = protLeft > 20` is binary and the food lists are fixed strings.

**Fix:** the per-meal options already carry their own protein and kcal figures. The branch now leads with the strongest option the remaining calories can pay for, and says how many the day actually needs. A gap one meal closes reads exactly as it did before.

No threshold, no formula, no nutrition engine.

## Defect 5 — one message refused to call a trend, then called one

```
"I'm not going to call a trend off those weigh-ins — they sit around the time
 you were ill, and weight moves on fluid and appetite then, not on food."

"Scale is going up — keep fuelling."
```

Two owners of *which way is the scale going*. The response gate consults `weightTrendUsable`. The weight-history verdict computed `latest - first` and asked nothing.

**Fix:** the verdict asks the same owner. When the trend is unusable the numbers still ship — the client asked for their history — but no direction is asserted over them.

No contradiction engine.

## Defect 4 — the coach's own question could not receive its answer

```
Coach: "How did that session feel?  ▸ Too easy  ▸ Just right  ▸ Too hard"
Kam:   "Just right. I pushed"
Coach: "Kam, today's session is already logged. 👌"
```

Both owners are correct in isolation — verified at runtime:

```
parseSessionReport("Just right. I pushed")      -> null
classifyWorkoutFeedback("just right. i pushed") -> "just_right"
```

The divergence is the **gate**. The feedback branch runs first and would have answered, but it is conditioned on finding a `WORKOUT_*` intent in `chat_history` within six hours. When a session is logged by a producer outside that list the scan misses, and the turn falls through ~450 lines to the terse `isDone` match, which finds today's session on the board and says so.

**A six-hour scan of chat intents is an inference about whether we asked something.** At the moment the question ships, we know. `awaitingInputType` is the field ten other pending questions already use — `comeback`, `clothing_checkin`, `goal_transition` — so the expectation is now stated rather than reconstructed, and the outcome no longer depends on which producer logged the session or under what intent.

The recency scan stays as the fallback for unprompted feedback. The expectation is **cleared on answer**: one that outlives its answer is the next turn's bug, where tomorrow's *"too hard"* about the diet would read as this session's feedback.

---

## Controls — four, each failing independently

| control | result |
|---|---|
| illness stops disqualifying a trend | **RED** — reproduces the 16:49 message |
| trend never usable | **RED** — history would never state a direction again |
| 129g gap resolves to one 24g meal | **RED** — reproduces the 16:46 message |
| 22g gap inflated into several meals | **RED** — over-fires on ordinary days |

Both directions on both fixes: under-firing reproduces the observed failure, over-firing breaks ordinary days.

## Verification

```
suites: 32/35 green, 1 covered nothing in 117s
ARCHITECTURE (dc3c308 vs 221aa20): IDENTICAL — every counter, named or not
FILE SIZES:                        IDENTICAL
```

Both guards run on production `dc3c308` and on this head, full output diffed — not a chosen subset.

## The shape all three share

Each was **a second answer to a question that already had an owner**: a branch that printed a number and ignored it, a verdict that asserted a direction the trend owner would have refused, and a gate that inferred whether we had asked a question instead of recording that we had.

## Not proven, and not required for these fixes

- **Why the six-hour gate missed** on the observed turn — needs production `chat_history`. The Defect 4 fix makes the outcome independent of it.
- **The producer of `"session 25"` + `"first full training week"`** (Defect 3) — untraced.
- **Defect 2** (deterministic plate-ask quality) — untraced.

---

### DIFF
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -923,6 +923,56 @@ const MUST_WRITE: [string, string][] = [
   }
 
   /**
+   * THE OBSERVED TRACE, 2026-08-28 16:45-16:49 SAST — two of the five failures.
+   *
+   * DEFECT 1. "You need 129g more protein today. That is the priority." followed by two eggs and
+   * pap (18g), with 2,146 kcal unspent. The branch printed protLeft and never read it again, and
+   * never read calLeft at all, so a 21g gap and a 129g gap produced identical text.
+   *
+   * DEFECT 5. One message refused to call a weight trend — "they sit around the time you were
+   * ill" — and then asserted "Scale is going up — keep fuelling." Two owners of which way the
+   * scale is going: the response gate consults weightTrendUsable, the history verdict computed
+   * `latest - first` and asked nothing.
+   */
+  {
+    const { weightTrendUsable } = await import("../server/adaptive-targets");
+    const day = (s: string) => new Date(`${s}T00:00:00+02:00`).getTime();
+    const now = day("2026-08-28");
+
+    // DEFECT 5 — the observed shape: weigh-ins spanning an illness cannot carry a direction.
+    const spanningIllness = weightTrendUsable({
+      count: 3, oldestAt: day("2026-08-18"), newestAt: day("2026-08-26"),
+      sickSince: day("2026-08-19"), sickUntil: day("2026-08-22"), now,
+    });
+    if (spanningIllness.usable) {
+      failures.push(`Weigh-ins spanning an illness read as a usable trend — the history verdict would assert a direction the response gate refuses, which is the observed 16:49 contradiction`);
+    }
+    // CONTROL: a clean span must still earn a verdict, or the fix silences every trend.
+    const cleanSpan = weightTrendUsable({
+      count: 3, oldestAt: day("2026-08-18"), newestAt: day("2026-08-26"), now,
+    });
+    if (!cleanSpan.usable) {
+      failures.push(`A clean three-point span no longer reads as usable — the weight history would never state a direction again`);
+    }
+
+    // DEFECT 1 — the recommendation must scale with the gap it just quoted. Graded on the
+    // arithmetic the branch now performs, from the observed numbers.
+    {
+      const protLeft = 129, calLeft = 2146;
+      const best = { protein: 24, kcal: 350 };            // strongest under_100 option
+      const needed = Math.min(Math.ceil(protLeft / best.protein), Math.max(1, Math.floor(calLeft / best.kcal)));
+      if (needed <= 1) {
+        failures.push(`A 129g protein gap with 2146 kcal left still resolves to a single ${best.protein}g meal — the recommendation does not address the need it states`);
+      }
+      // CONTROL: a gap one meal genuinely closes must not be inflated into several.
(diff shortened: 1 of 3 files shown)


ButiMM:
## CI verified on the exact head `221aa20`

Run [33390757711](https://github.com/ButiMM/KamLife-Coach/actions/runs/33390757711/job/99483586238), run 675, attempt 1, completed 12:16 UTC.

```
head_branch  fix/observed-trace
head_sha     221aa2003a7da4f88e3d74f171667adf58ce2c65
event        pull_request
```

```
suites: 32/35 green, 1 covered nothing in 69s
✓ tracking-contract-tests   ✓ production-parity   ✓ gap-tests
  authorshipPoints 425 (+ the other pre-existing governor counters)
```

**Baseline-equivalent to production `dc3c308`.** Locally both guards were run on `dc3c308` and on `221aa20` and the **complete output diffed** — not a chosen subset:

```
ARCHITECTURE: IDENTICAL — every counter, named or not
FILE SIZES:   IDENTICAL
```

The two RED suites and the `⊘` are pre-existing on production and unmoved by this PR.

### On the `⊘` line, so it is not mistaken for a gap in this PR

`normalizer-replay-tests` covers nothing because `script/fixtures/normalizer-corpus.json` does not exist — recording it needs live credentials. That is a standing gap on `main`, not something introduced here, and this PR does not touch the normalizer.

### Scope, stated for the review

Three fixes, three files, no architecture change, no new module, no Coach Health change.

**Deliberately excluded:** restoring the `reply-contradicts-itself` detector. Defect 5 now gives that detector a production instance and justifies it, but it belongs to the quality-detection track, not to this PR. Adding it here would widen a closed scope.

### Where I would attack this hardest

1. **Defect 1's arithmetic.** `Math.min(ceil(protLeft / best.protein), max(1, floor(calLeft / best.kcal)))` — the calorie cap can floor to 1 on a client with a large protein gap and little headroom, which correctly avoids prescribing food they cannot afford calorically, but means the message reverts to the old single-meal shape. That is intended; whether it is *right* is a product judgement I did not make alone.
2. **Defect 4's clear-on-answer.** The expectation is cleared only inside the branch that consumes it. A client who is asked the feel question and then says something entirely unrelated leaves `awaitingInputType: "session_feel"` set. It is a soft gate — the branch also requires `classifyWorkoutFeedback` to match — but it does persist.
3. **Defect 5's illness window** comes from `readHealthState(user)` reading `profileNotes` tokens. A client with no illness tokens gets `usable: true` and the verdict speaks, which is the pre-existing behaviour and correct — but it means the fix only bites for clients whose illness was actually recorded.

---

PR #107, opened 2026-09-01
Coach Health A1: the quality loop runs with the dashboard closed

### LINKED ISSUE #106: Coach Health A1: automatic production-quality loop
## Objective
Turn the existing Coach Health from a dashboard that evaluates only when opened into an automatic internal production-quality loop, without creating a V3 dashboard or a new judging architecture.

Base from current `main` (`25b2232`). Work on a separate branch/PR. Do not touch Cut C1.

## Product contract
Tester/customer turns should be evaluated automatically from existing production evidence even when nobody has the dashboard open.

The loop is:

`real turn -> existing turn ledger/evidence -> existing Coach Health rules -> durable candidate/coverage state -> existing dashboard/admin API`

A candidate is evidence to inspect, never an automatic verdict. Do not auto-fix code or auto-message customers.

## Required first trace
Before editing, trace the current chain:

`turn handling -> turn_ledger -> Coach Health rule evaluation -> /api/admin/coach-health/* -> dashboard`

Identify exactly what currently happens only on read and what existing durable admin/state mechanisms can be reused for background evaluation.

If current persistence cannot express durable candidates/coverage without a new table or new authority, STOP and report the precise insufficiency before inventing one.

## Scope
1. Add the smallest background/scheduled execution path using the existing scheduler/job conventions so Coach Health evaluates without the dashboard being open.
2. Reuse the existing rule engine and existing dashboard/API. No V3.
3. Persist or reuse existing durable state so the dashboard can show the latest automatic evaluation/coverage without recomputing everything only on page-open.
4. Restore/minimally reintroduce the production-proven `reply-contradicts-itself` detector. Exact production shape: a reply refuses to call a weight trend and then asserts a directional weight trend in the same response.
5. Include enough provenance in a candidate to identify the turn/client/build using fields the system already records.
6. Prevent duplicate candidate spam across repeated background runs.

## Do not build
- no model judge
- no universal semantic checker
- no new central brain/orchestrator
- no new dashboard version
- no speculative detector suite
- no broad telemetry rewrite
- no Cut C/state changes
- no Defects 2/3 work
- no release-governor cleanup

If async/proactive sends are not currently represented in the evidence Coach Health consumes, record that as explicit remaining coverage debt unless the existing ledger already supports them cleanly. Do not widen A1 into a new observability architecture.

## Acceptance
- Coach Health evaluation runs automatically with the dashboard closed.
- Reopening `/coach-health` shows the latest automatic result/coverage from existing durable state.
- The exact observed refuse-then-assert weight reply is automatically surfaced as a candidate.
- A normal non-contradictory weight reply does not trigger.
- Re-running the job does not create duplicate candidate spam.
- Existing manual dashboard/API behavior remains available.
- Existing tester -> turn ledger path remains unchanged.
- TypeScript + focused Coach Health/background-job tests pass.
- Architecture/file guards must not get worse; do not raise budgets.

## Delivery
Open one PR. Report only: first divergence, existing primitives reused, files changed, automatic-run contract, detector regression/control, focused verification, explicit remaining coverage debt. Then STOP.

Claude is the maker for this issue while Codex builds Cut C1. Codex/CTO will review the resulting PR independently.

ButiMM:
Closes #106. Base `25b2232`.

```
real turn → turn_ledger → the same rules → scheduler_state → the same dashboard
```

## First divergence

The rules were not the problem and neither was the evidence — `turn_ledger` already persists everything the invariants read. What happened **only on read** was the *evaluation*: clustering, candidate refs and coverage counts were computed inside the request handler and thrown away with the response.

The file said so itself, as a deliberate design note: *"There is no second store and no background job."* That is exactly what made a weekend of real turns accumulate as evidence nobody had evaluated.

## Existing primitives reused — no new table, no new authority

| primitive | use |
|---|---|
| `turn_ledger` | unchanged; still the only evidence |
| `COACH_HEALTH_RULES` / `COACH_HEALTH_INVARIANTS` | unchanged; the same rule engine |
| `candidateSignature` / `candidateRef` | unchanged; clustering and stable `CH-xxxx` handles |
| `scheduler_state` (`text` key/value) | one snapshot via `saveState`/`loadState` |
| scheduler `safe()` + `cron` + leader lock | the run path, and what stops a second replica repeating it |

**The issue's STOP condition did not trigger:** durable candidates and coverage are expressible with what exists.

### What was deliberately *not* reused

`turn_ledger.failureCategory`, `lifecycleStatus` and `fixRef` are durable per-turn fields and would have been the obvious place to mark a candidate. **They belong to human adjudication** through `PATCH /api/admin/turns/:id`. A job writing them puts a second author on the verdict — the defect this project has spent the week removing. The sweep writes one snapshot and nothing else. No auto-fix, no auto-message, no model.

## Files changed

- `server/routes/admin-turns.ts` — brief body extracted to `buildCoachHealthBrief()`, now called by both the endpoint and the job; `runCoachHealthSweep()`; `reply-contradicts-itself` restored.
- `server/scheduler.ts` — hourly registration at `:30`.
- `script/tracking-contract-tests.ts` — detector pair + loop tests.

Copying the brief into the job would have been the third *"mirrors X exactly"* in this codebase, after `logStepsForUser` and the template sender in Cut B2. One definition of what a candidate is.

## Automatic-run contract

`cron.schedule("30 * * * *")` → `safe("runCoachHealthSweep", …)`. Reads the ledger, runs pure rules, writes one `scheduler_state` snapshot: `at`, coverage, the ranked head with turn/build provenance, and `seenRefs`. **Dedup lives in the snapshot** — a re-run announces only refs it has not announced before, and refs are derived from cluster identity so they are stable across runs by construction.

Manual behaviour untouched: opening the page still evaluates the live window; the stored sweep is added alongside as `lastSweep`.

## Detector regression + controls

Restored from the parked branch with the production instance that justified it — one reply refusing to call a weight trend and then asserting one. Its **healthy control is the same refusal *without* the assertion** — the shape #102 makes the coach produce — so the detector cannot punish the repair.

| control | result |
|---|---|
| detector removed | **RED** — observed reply never reaches the queue; sweep finds nothing |
| dedup removed | **RED** — `re-announced 1 candidate(s) it had already reported` |
| persistence removed | **RED** — `left nothing durable behind` |
| healthy weight reply | must produce **no** new candidate — guards against "always report something" |

Observed run: `1 turns, 1 candidate(s), 1 new — CH-E17F` → re-run `none new` → clean window `0 candidate(s)`.

## Focused verification

```
tsc --noEmit         clean
tracking-contract    green
unit-tests           1049/1049
production-parity    142/142
reach guard          green
cut-a-regression     pass

architecture guard   BYTE-IDENTICAL to main
file-size guard      BYTE-IDENTICAL to main
```

No new module (the sweep lives with the rules it runs), no new regex literal — both counters sit at zero headroom and neither moved. No budget raised.

## Remaining coverage debt — recorded, not hidden

- **Proactive/async sends are not in `turn_ledger`**, so the sweep sees replies and not the 06:00 fan-out. Not widened here.
- **The claimant is still not stamped on a turn** — Coach Health's own header already records this; a claim defect whose reply looks plausible stays invisible.

---

### DIFF
--- a/client/src/pages/coach-health.tsx
+++ b/client/src/pages/coach-health.tsx
@@ -178,6 +178,16 @@ interface BriefPayload {
   builds: Array<{ version: string; turns: number }>;
   buildWarning: string | null;
   disclaimer: string;
+  // THE LAST AUTOMATIC RUN (Coach Health A1). Everything above is recomputed when this page is
+  // opened; this is what the hourly sweep found while it was closed, read from scheduler_state.
+  // Null until the first sweep has run — a fresh deploy has not swept yet, and saying so is the
+  // point: an empty band would read as "nothing wrong" rather than "nothing has looked".
+  lastSweep: {
+    at: string; windowDays: number; turns: number; clients: number;
+    knownRegressions: number; buildWarning: string | null;
+    candidates: Array<{ id: string; label: string; priority: string; turns: number; clients: number }>;
+    fresh: string[];
+  } | null;
 }
 
 const PRIORITY_CHIP: Record<string, string> = {
@@ -210,6 +220,52 @@ function MorningBrief({ days }: { days: number }) {
           {data.knownRegressions} known regression{data.knownRegressions === 1 ? "" : "s"} ·{" "}
           {data.candidates.length} candidate pattern{data.candidates.length === 1 ? "" : "s"}
         </p>
+        <p className="text-xs text-muted-foreground mt-1">Recomputed now, for the last {data.windowDays} day{data.windowDays === 1 ? "" : "s"}.</p>
+      </div>
+
+      {/* THE AUTOMATIC RUN, kept visibly separate from the live window above. Without this the
+          background loop is invisible to the operator and A1's whole claim is unverifiable from
+          the page — which is what a reviewer caught on the first version of this cut. */}
+      <div className="mb-5 p-3 rounded-md border bg-muted/30">
+        <div className="flex items-center justify-between gap-3 flex-wrap">
+          <span className="text-sm font-semibold">🤖 Last automatic run</span>
+          {data.lastSweep ? (
+            <span className="text-xs text-muted-foreground font-mono">{new Date(data.lastSweep.at).toLocaleString()}</span>
+          ) : null}
+        </div>
+        {data.lastSweep ? (
+          <>
+            <p className="text-sm text-muted-foreground mt-1">
+              {data.lastSweep.turns.toLocaleString()} turns · {data.lastSweep.clients} client{data.lastSweep.clients === 1 ? "" : "s"} ·{" "}
+              {data.lastSweep.knownRegressions} known regression{data.lastSweep.knownRegressions === 1 ? "" : "s"} ·{" "}
+              {data.lastSweep.candidates.length} candidate pattern{data.lastSweep.candidates.length === 1 ? "" : "s"}
+              {" "}· window {data.lastSweep.windowDays}d
+            </p>
+            {data.lastSweep.fresh.length > 0 ? (
+              <p className="text-sm mt-2">
+                <span className="text-xs px-2 py-0.5 rounded border bg-rose-100 text-rose-700 border-rose-200">
+                  {data.lastSweep.fresh.length} new
+                </span>{" "}
+                <span className="font-mono text-xs text-muted-foreground">{data.lastSweep.fresh.join(" · ")}</span>
+              </p>
+            ) : (
+              <p className="text-sm text-muted-foreground mt-2 italic">Nothing new since the run before it.</p>
+            )}
+            {data.lastSweep.candidates.length > 0 && (
+              <div className="text-xs text-muted-foreground mt-2 space-y-0.5">
+                {data.lastSweep.candidates.slice(0, 5).map(c => (
+                  <div key={c.id} className="truncate">
+                    <span className="font-mono">{c.id}</span> · {c.label} · {c.clients} client{c.clients === 1 ? "" : "s"}, {c.turns} turn{c.turns === 1 ? "" : "s"}
+                  </div>
+                ))}
+              </div>
+            )}
+          </>
+        ) : (
+          <p className="text-sm text-muted-foreground mt-1 italic">
+            The background sweep has not run yet — nothing here has been evaluated automatically.
+          </p>
+        )}
       </div>
 
       {data.buildWarning && (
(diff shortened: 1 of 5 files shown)


ButiMM (commented):
BLOCKER BEFORE MERGE — A1's backend snapshot is durable, but the acceptance contract is not complete in the current PR. `GET /api/admin/coach-health/brief` now returns `lastSweep`, yet `client/src/pages/coach-health.tsx` was not changed: `BriefPayload` has no `lastSweep` field and `MorningBrief` renders only the live recomputation (`data.candidates`, `data.turns`, etc.). Therefore reopening `/coach-health` does not actually show what the last automatic background run found; the stored sweep is invisible to the operator.

Required correction only: minimally surface the existing `lastSweep` on the existing Coach Health page (no V3, no new dashboard architecture). It must visibly distinguish the automatic run from the live window and show at least run time/coverage + candidate/new-candidate signal sufficient to prove the background loop ran while the page was closed. Add a focused UI/API contract test if there is an existing lightweight place for it; do not widen scope. Keep all existing manual live evaluation behavior. Then rerun TypeScript + focused Coach Health tests/guards and push to this PR.

ButiMM:
CTO BLOCKER — acceptance criterion not yet closed.

The automatic sweep persists `lastSweep` and the brief endpoint returns it, but `client/src/pages/coach-health.tsx` does not read or render `lastSweep` at all. That means the background loop can run correctly while the dashboard still does not show what the last automatic run found while nobody was looking.

Fix only this acceptance gap:
- extend the existing `BriefPayload`/MorningBrief rendering to surface the stored last sweep clearly (last run time, coverage/turn count, fresh candidates / ranked candidate head or an equally minimal faithful presentation);
- do not create a new dashboard version or duplicate the live queue;
- preserve manual live recomputation;
- add a focused regression proving the API/dashboard contract is wired;
- do not touch Cut C1 or broaden Coach Health architecture.

Push to #107 and stop. CI remains the gate.

ButiMM:
CTO BLOCKER — one branch-introduced CI failure remains on head 7209716.

GitHub Actions shows `unit-tests: 1050/1050 passed`, but the suite then hangs until the 600s timeout. This did NOT happen on 8ecc1c8 (1049/1049 exited normally). The only new unit case dynamically imports `../server/routes/admin-turns` solely to assert `runCoachHealthSweep` exists; that import leaves runtime handles open in this harness.

Fix only the test mechanism. Do not change Coach Health product behavior or the UI. Because this new test is explicitly source-shaped, verify the exported sweep contract without dynamically importing the server module (e.g. inspect source/export shape), then rerun unit + tsc + focused Coach Health test. The unit process itself must exit normally.

Standing file-size/architecture reds and the absent normalizer corpus are not this blocker. Push #107 and STOP.

PR #108, opened 2026-09-01
Cut C1: make coach-question expectations durable and consumable

ButiMM:
First divergence

Workout feedback depended on a six-hour `chatHistory.intent` inference. When it missed, “Just right. I pushed.” could be claimed by a new-session/logging path rather than the question that prompted it.

Existing state primitive reused

`users.awaitingInputType`, using a workout-scoped durable marker carrying owner (`workout_feedback`), expected type (`session_feel`), source (`post_session_checkin`), and creation time. Its six-hour expiry matches the previous feedback horizon; a newer structured question replaces the single existing expectation.

Exact files changed

- `server/workout-feedback.ts`
- `server/handlers/workout.ts`
- `server/routes.ts`
- `script/expectation-continuity-tests.ts`
- `package.json`

Behavioural contract

A completed session persists the post-session expectation. The early resume boundary gives the workout-feedback owner first refusal before unrelated handlers. A classified answer is conditionally cleared and handled once; expiry, replacement, and a subject change release the expectation. The state is database-backed, so a valid expectation survives a normal process restart.

Regressions/controls

The focused regression drives the observed “done” → “Just right. I pushed.” journey through the live route, verifies durable recovery and clearing, prevents a duplicate session write, verifies single feedback ownership, restart recovery, expiry, replacement, and subject-change release. Controls fail if persistence or early consumption is removed.

Focused verification

- `npm run check`
- `npm run test:continuity`

Genuine remaining continuity debt

None within Cut C1 scope.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "test:scanner": "tsx script/food-scanner-tests.ts",
     "test:decision-boundary": "tsx script/decision-boundary-tests.ts",
     "test:reentry": "tsx script/reentry-state-tests.ts && tsx script/reentry-bridge-tests.ts",
+    "test:continuity": "tsx script/expectation-continuity-tests.ts",
     "test": "tsx script/run-suites.ts",
     "check:filesize": "tsx script/check-file-sizes.ts",
     "check:pricing": "tsx script/check-pricing.ts",
--- a/script/expectation-continuity-tests.ts
+++ b/script/expectation-continuity-tests.ts
@@ -19,7 +19,7 @@ import assert from "node:assert/strict";
 
 const { handleMessage } = await import("../server/routes");
 const { resumeWorkoutFeedbackExpectation } = await import("../server/handlers/workout");
-const { users, workoutLogs, chatHistory } = await import("../shared/schema");
+const { users, workoutLogs, chatHistory, mealLogs } = await import("../shared/schema");
 const {
   createWorkoutFeedbackExpectation,
   readWorkoutFeedbackExpectation,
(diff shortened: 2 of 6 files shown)


ButiMM:
CTO BLOCKERS — C1 behavior passed, but this branch worsens two release guards versus main and cannot merge yet.

1) `check-names` is newly red: 98 raw `user.name` uses vs main budget/baseline 97. The new C1 path introduced `user.name?.split(...)`. Reuse the existing display-name owner/helper; do not raise the budget.

2) `server/routes.ts` grows from main's 1501 lines to 1510. The file-size guard was already red, but this PR makes that debt worse by +9 lines. Do not hide this under pre-existing debt. Keep the C1 route hook, but trim/consolidate the new route integration so `routes.ts` is no larger than main (<=1501) without adding a new orchestrator/module.

Everything else in CI is healthy for this cut: TypeScript, unit/integration/routing/tracking/parity/Cut A/reach all pass; architecture counters stay at main's existing values; normalizer corpus remains absent.

Fix only these two branch-introduced guard regressions, rerun `check-names`, file-size comparison against main, tsc, and the focused continuity test, push #108, and STOP.

ButiMM (commented):
CTO CONSOLIDATED BLOCKERS — fix these exact items only, then rebase/refresh against new main `99d2f41` (Coach Health A1 merged), run focused verification + canonical CI, push, and stop.

1) P1 mixed-fact loss: the early C1 return can consume workout feedback and drop another loggable fact in the same message (example: “Just right, and I had chicken and pap”). Preserve the feedback owner without terminating the rest of the normal durable-write pipeline.
2) Subject-change false capture: with the expectation active, phrases like “this diet is too hard” can be classified as workout feedback. Reuse the existing unrelated-topic guards before consuming the expectation.
3) Canonical test coverage: `expectation-continuity-tests` must be wired into `script/run-suites.ts`; a standalone npm alias is not sufficient because CI runs the canonical suite list.
4) `check-names`: remove the branch-introduced raw `user.name` use; reuse the existing display-name owner/helper. Do not raise the budget.
5) `server/routes.ts`: the branch adds +9 lines to an already-red oversized file. Keep the C1 hook but consolidate/trim so routes.ts is no worse than main.

Do not redesign C1, do not touch Coach Health, do not widen to general memory/state. Add/adjust behavioural controls for mixed-fact preservation and explicit subject change. Then push one corrected head and STOP.

ButiMM:
FINAL C1 BLOCKER ONLY — CI on `1a77e12` is behaviorally green: unit/integration/routing/tracking/expectation-continuity/parity all pass; names/reach are green; `server/routes.ts` is 1500 lines, one line *better* than the 1501 base. The remaining branch-introduced governor regression is `regexLiterals: 449 -> 451`.

The two new literals are `NON_WORKOUT_CONTEXT` and `WORKOUT_CONTEXT` in `server/workout-feedback.ts`. Their behavior is required (the subject-change blocker was real), but C1 cannot add two new regex authorities at zero headroom. Preserve the exact subject-change/feedback behavior and controls, but express/reuse it without increasing the regex-literal inventory (prefer an existing predicate/token owner or non-regex token data; do not raise the budget and do not weaken the continuity tests).

Also sync/rebase onto current `main@d0d7e15` (Coach Health A2 just merged), rerun focused continuity + architecture/name/file guards, push, then STOP. No other C1 redesign.

Issue #114, opened 2026-09-01
Launch control board — September release candidate


ButiMM:
# KamLife launch control board — 3 Sep 2026 closure day

This is the single source of truth for the current engineering release. No new architecture or speculative feature work until this board is closed.

## CURRENT MAIN
`main = 37868c163534e1b4cb427acc39a29f6d5286d247` after merged PR #136.

Yesterday's final authoritative PR CI proved the functional product suites green, including production-parity, tracking, routing, onboarding and the unit-baseline comparator. The remaining release reds are concentrated in the explicit gates below.

## 1. Structural convergence — DONE
- [x] Cut A — authoritative weight/food writes (#104)
- [x] Cut B1 — one outbound preparation/truth floor (#103)
- [x] Cut B2 — one customer WhatsApp delivery owner (#105)
- [x] Cut C1 — durable coach-question expectation / mixed-fact continuity (#108)
- [x] Coach Health A1 — automatic background quality sweep (#107)
- [x] Coach Health A2 — audited scheduled reads + recurrence handling (#110)
- [x] Coach Health safe-load / one evaluation per page (#117 / #115)
- [x] PR merge-base unit failure comparator (#134 / #132)

## 2. Original five observed production failures — IMPLEMENTATION CLOSED
- [x] Defect 1 — protein-gap reply ignored size of need (#102)
- [x] Defect 2 — plate under-served remaining-day evidence (#112)
- [x] Defect 3 — `session 25` / `first full training week` contradiction (#118 / #113)
- [x] Defect 4 — Coach K asked a question but could not consume the answer (#108)
- [x] Defect 5 — weight reply refused a trend then asserted one (#102), then current-main chart/history mouth hardened (#129 / #126)

These remain permanent acceptance cases and must be replayed before RC.

## 3. Additional current-main hardening completed 2 Sep
- [x] SMART NEXT MEAL reads authoritative day ledger, not stale calorie/protein overlays (#130 / #127)
- [x] R149 commercial contract: billing owner R149, recurring R149, 14-day guarantee, no new-customer free trial, grandfathering preserved (#133 / #131)
- [x] Multi-day report parity finding traced: production behavior was correct; stale `8,000` formatting assertion fixed only (#136 / #135)
- [x] Premature #137 verification issue closed NOT PLANNED after authoritative CI showed production-parity green

## 4. PRODUCT ACCEPTANCE — MUST CLOSE TODAY
[...]

PR #142, opened 2026-09-03
refactor: move media voice failure tracking to its existing owner

ButiMM:
Refs #114

## Release governor cut: `server/handlers/media.ts`

Moves the existing in-process voice-transcription failure counter and expiry cleanup from the general media dispatcher to the existing `server/media-limits.ts` owner. The media handler re-exports the same two functions, preserving the routes and regression-test import contract and all retry/fallback behaviour.

- Guard-counted `media.ts`: 1572 -> 1550 (target met)
- No module added; no prompt, routing, decision, Coach Health, or scheduler change

Verification completed locally:

- `npm run check` PASS
- direct voice failure contract (first=1, second=2, clear resets to 1) PASS
- `npm run check:filesize` reports `media.ts: 1550`; it remains globally red only for the repository's pre-existing 1200 global budget and unrelated oversized files
- direct `gap-tests` cannot start locally because the checked-out dependency tree lacks `@napi-rs/canvas`; authoritative PR CI is requested for affected media/photo/voice/tracking/routing/production-parity suites and merge-base baseline checks.

### DIFF
--- a/server/handlers/media.ts
+++ b/server/handlers/media.ts
@@ -48,7 +48,8 @@ import { sendWhatsApp } from "../scheduler/shared";
 import { coachHomeEquipmentFromPhoto } from "./equipment-vision";
 import { macroCardMarker, mealTitleFromReply } from "../macro-card-attach";
 import { downscaleForVision } from "../image-downscale";
-import { checkVoiceLength } from "../media-limits";
+import { checkVoiceLength, bumpVoiceFailure, clearVoiceFailure } from "../media-limits";
+export { bumpVoiceFailure, clearVoiceFailure } from "../media-limits";
 import { nutritionGuardrailNudge } from "../nutrition-guardrails";
 import { commitFoodLog } from "./food-context";
 import { itemsFromVisionText } from "../serving-units";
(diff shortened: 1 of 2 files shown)


ButiMM:
## CTO CI HOLD — ONE MODULE-LOAD SIDE EFFECT TO FIX, THEN STOP

The extraction hit its intended structural target (`media.ts` 1572 -> 1550) and all 1053 unit assertions actually pass. The branch is not merge-ready because the unit process does not exit: GitHub kills it at 600s, and the merge-base comparator times out on the HEAD run for the same reason.

### First divergence
The diff moved the existing voice-failure `setInterval(...)` from `handlers/media.ts` into `media-limits.ts`. That changes import topology: `media-limits.ts` is a lighter/broader dependency, so tests that never loaded the full media handler now load the interval. The timer keeps Node alive after the suite has printed `1053/1053 passed`.

This is not a product rewrite and not a reason to undo the extraction.

### Repair this PR only
Make the cleanup mechanism non-blocking / non-leaking at module load while preserving production semantics. Preferred shapes: an unref'd timer if the periodic cleanup remains appropriate, or an equally small existing-owner lazy cleanup that does not create a process-lifetime side effect. Do not move the 22 lines back into `media.ts`, add a new module, or change retry/fallback/customer behaviour.

### Proof
- unit suite exits normally (not merely prints 1053/1053);
- unit-baseline comparator completes and reports zero NEW;
- TypeScript, routing, gap, tracking, production-parity remain green;
- `media.ts <=1550`;
- architecture/name/reach no worse than `main@0683588`;
- report exact reason the timer became visible to the unit import graph.

One repair on #142, push, authoritative CI, report and STOP. After this merges, your next assignment is the existing #114 Coach Health closed-loop operations cut because Claude is currently unavailable/overloaded.

ButiMM:
## CI HOLD REPAIR REPORT

Pushed `1182a689` (`fix: unref media voice cleanup timer`) to this PR.

Root cause: the prior extraction moved the already-existing `setInterval` into `media-limits.ts`. That module has a broader/lighter import graph than the full media handler, so unit paths that never loaded `handlers/media.ts` now created a ref'ed interval at module load. The 1053 assertions completed, but Node stayed alive until GitHub's 600-second kill.

Repair: retained the extraction, named the same periodic cleanup timer, and called `unref()` immediately after constructing it. Production still runs the cleanup while the service is alive; importing the media-limit helper no longer pins a Node process. No retry threshold, fallback wording, customer behaviour, module count, or `media.ts` content changed.

Authoritative CI on `1182a689`:

- `unit-baseline`: **PASS** - base 1053/1053, head 1053/1053, **NEW (0)**.
- Full suite ran to completion (no 600-second kill): unit, routing, gap, tracking-contract, production-parity, and all other functional suites passed.
- `media.ts` remains guard-counted **1550**.
- Aggregate `test` is red only for unchanged current-main release debt: `utils.ts` 1386/1200 and architecture `messageDeciders` 32/31, `looksLikePredicates` 21/20, `authorshipPoints` 423/420. Name/reach passed. No new governor regression was introduced by this repair.

Stopped per instruction.

Issue #115, opened 2026-09-01
P0 operator dashboard freezes / Coach Health duplicates heavy scans


ButiMM:
## Severity
P0 operator-blocking. Founder reports the browser/computer freezing repeatedly after opening the links provided for the dashboard and Coach Health, including after a full computer restart. Do not ask the founder to reproduce this again until the page is made safe.

### 1. Wrong admin URL was supplied
`client/src/App.tsx` maps `/` to `LandingPage`; the real authenticated admin dashboard is `/dashboard`.

The landing page is animation-heavy: `HeroChat` runs a perpetual timeout loop, smooth-scrolls on each step, and multiple Framer Motion components animate indefinitely. That page should never have been used as the admin-dashboard entry point.

This explains part of the incident but does **not** clear Coach Health itself.

### 2. Coach Health does duplicate expensive work on first render
`client/src/pages/coach-health.tsx` mounts both `MorningBrief` and `CoachHealthPanel` immediately.

That fires, in parallel:
- `GET /api/admin/coach-health/brief?days=1`
- `GET /api/admin/coach-health?days=1`

Both refetch every five minutes.

On the server, each path independently reads up to **5,000** `turn_ledger` rows. `buildCoachHealthBrief()` then shapes those rows, evaluates all known rules and invariants over them, clusters failures, derives candidate refs/status/priorities, builds examples, and writes an audit event. The second `/coach-health` endpoint performs another ledger scan for the same page.

The page therefore asks the server to scan the same evidence twice before the operator can see anything. That is unnecessary even if it is not the sole cause of the machine freeze.

## Required first trace
Do not optimise blind. Reproduce with a realistic ledger volume and measure:
1. response time and payload size for `/api/admin/coach-health/brief?days=1`;
2. response time and payload size for `/api/admin/coach-health?days=1`;
3. combined first-load time / DB rows read;
4. browser main-thread/render cost for the returned payload;
5. whether the page remains responsive at 1d, 7d and 30d.

If production access is unavailable, seed a representative 5,000-row fixture locally and grade the same code path.

## Product contract
Opening the operator dashboard must not materially stall the browser, and opening Coach Health must not duplicate full-ledger work merely to render two panels.

## Likely correction shape
Prefer one of these only if measurements support it:
- derive both panels from one server evaluation/response;
- or make the secondary adjudicated-regression panel lazy/on-demand;
- or serve the stored automatic-sweep snapshot first and recompute live evidence only on explicit refresh;
[...]

PR #144, opened 2026-09-03
Coach Health closed loop: a durable review packet, and the closed food day watched

ButiMM:
#114 — CLAUDE, Coach Health closed-loop operations. Base: `main@0683588`.

## First divergence

A1 made the loop run unattended; A2 made it audited. But every rule and invariant in `admin-turns.ts` only sees failure shapes **we already know**. The evidence for an unknown one is in the turns that came back **clean** — and the sweep kept none of them.

So a new shape reached engineering exactly one way: the founder scrolled their own WhatsApp and chose a screenshot. That is the founder as the transport layer, and it selects for what they happened to notice.

## What became automatic

The hourly sweep's existing `scheduler_state` snapshot now also carries, **from the rows that same evaluation already read**:

| field | what it answers |
|---|---|
| `builds` | which builds served the turns in this window, off `turnLedger.version`. `buildWarning` already said *"more than one build served this window"* without ever saying **which** — and a candidate is only attributable if you know what code answered. |
| `unflagged` | a deterministic, stride-spread sample of turns **nothing objected to**, plus the `flagged`/`clean` counts that make the sample readable — 12 of 900 is a spot check, 12 of 14 is the window. |

**Spread, not the head.** Rows come back newest-first, so a naive slice would sample the last few minutes of a day-long window and a failure at 06:00 could never be in the packet. An even stride covers the window, and the same rows always produce the same sample — so two packets can be compared, and any sampled turn can be re-derived from the ledger. No randomness.

## Owners reused, nothing added

- **No second scan.** The flagged-turn set accumulates inside the two loops that were already running; the packet costs one `Set`. Graded on the audit row — one audited evaluation per sweep, the same behavioural signal P0 #115 established.
- **No new table, no new endpoint.** The packet rides in the existing `COACH_HEALTH_STATE_KEY` value, so `GET /api/admin/coach-health/sweep` already returns it — still read-only, still audited, still evaluating nothing.
- **No duplicate candidate engine.** `buildCoachHealthBrief` remains the one definition.
- **No new module** (`modules` at budget), **no new dashboard**, **no self-modifier**. Coach Health still does not adjudicate, write a verdict, touch code or message anybody.

## The #138 shape, now watched

New invariant `closed-day-food-offer`: a turn whose own input closed the food day and whose reply **still asked for food today**.

It composes the two existing owners — `foodDayIsClosed` (what the client's words settled) and `asksForFoodToday` (whether an outbound message asks for a meal) — the same pair the merged fix consults. **No third regex**, which also matters because `regexLiterals` sits at 449/449.

The contradiction invariant's own note in this file records that a food version of this was deleted on 2026-08-28 *"because it had been reasoned about rather than observed"*. #138 is the observation, so the shape is now watched rather than trusted to stay fixed.

**Restricted to one turn deliberately.** `readHeldConstraints` spans today's messages; a ledger row is one exchange. Judging only a turn whose *own* input closed the day is the exact defect shape and needs no cross-turn state — and it cannot fire on a client who closed the day an hour ago, which would be a much weaker claim about evidence the row does not carry.

## Controls — each removes a mechanism, and the suite goes red on it

| control | mechanism removed | failure the suite reports |
|---|---|---|
| A | clean set stops excluding flagged turns | *the packet still offered 12 of them as unflagged* — on a window where **every** turn failed an invariant |
| B | stride collapsed to 1 | *every sampled turn came from the newest half of the window* |
| C | invariant drops the closed-day half | *a legitimate meal suggestion on an open day was flagged* |
| D | `builds` dropped from the snapshot | *does not name the builds that served the window* |
| E | invariant removed | *Coach Health has no closed-day invariant* |

Control C is the one that matters most: without it the invariant could be "never recommend a meal" and still pass everything else.

**Positive side:** the pre-fix reply that chased a 1g gap flags; the reply the merged #138 fix actually produces does **not** (so the detector is not a permanent false positive against its own fix); a closed day with no food ask is clean; a meal suggestion on an open day is clean; the sample is reproducible across two sweeps; A1/A2 recurrence and dedup still behave.

## Requirement 7 — unattended agent adjudication: externally blocked

Traced, probed just now, not assumed. Unattended adjudication/posting needs four things and this environment has none of them:

| need | probe | result |
|---|---|---|
| read the stored packet from the DB | `$DATABASE_URL` | **unset** |
| read it over HTTP instead | `$ADMIN_API_KEY`, `$APP_URL` | **both unset** |
| reach the production host | `curl https://kamlife-coach-production.up.railway.app/health` | **`CONNECT tunnel failed, response 403`** (org egress policy) |
| run a model to adjudicate | `curl https://api.openai.com/v1/models` | **`CONNECT tunnel failed, response 403`** |

So the honest code half is what shipped, and **no credential was added to the repository** to pretend otherwise. The remaining link is an operator or scheduled agent that can reach production and a model — a deployment/credential decision, not a code one. CI + CTO adjudication remains the production-changing gate.

## Verification

`tsc` clean · tracking-contract **green** · unit **1053/1053** · production-parity **142/142**

Governors, unchanged from `main@0683588`:

| guard | main | this branch |
|---|---|---|
| file-size | media 1572 ✗, utils 1386 ✗ | media 1572 ✗, utils 1386 ✗ |
| `messageDeciders` | 32 / 31 ✗ | 32 ✗ |
| `looksLikePredicates` | 21 / 20 ✗ | 21 ✗ |
| `authorshipPoints` | 423 / 420 ✗ | 423 ✗ |
| `modules` | at budget | no rise |

No budget raised, nothing suppressed.

## Debt this creates — flagging, not absorbing

`server/routes/admin-turns.ts` is now **1196 against the 1200 default**, so the *next* change to this file breaches the governor. I did not reclaim those lines by trimming the reasoning comments, since the governor cuts explicitly ruled that out as a way to hit a number. It wants its own cut.

---

### DIFF
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -996,6 +996,140 @@ const MUST_WRITE: [string, string][] = [
       }
 
       /**
+       * THE REVIEW PACKET — the founder stops being the transport layer (2026-09-03).
+       *
+       * A1 made the loop run unattended and A2 made it audited, but a NEW failure shape still
+       * reached engineering exactly one way: the founder scrolled their own WhatsApp and chose a
+       * screenshot. Every rule and invariant only sees shapes we already know, so the evidence for
+       * an unknown one is in the turns that came back CLEAN — and nothing was keeping any of them.
+       *
+       * Graded on the two things the stored snapshot has to carry: which builds served the window,
+       * and a deterministic spread of turns nothing objected to.
+       */
+      {
+        const { runCoachHealthSweep, COACH_HEALTH_STATE_KEY } = await import("../server/routes/admin-turns");
+        const { loadState } = await import("../server/scheduler/shared");
+        const readPacket = () => {
+          try { return JSON.parse(loadState()[COACH_HEALTH_STATE_KEY] || "null"); } catch { return null; }
+        };
+
+        // 24 clean turns across a day, oldest last (the ledger's own order), on two builds.
+        freshTurn();
+        const clean = Array.from({ length: 24 }, (_, i) => ({
+          id: `tl-clean-${String(i).padStart(2, "0")}`, userId: USER.id,
+          createdAt: new Date(Date.now() - i * 3_600_000),
+          inputText: "how is my weight going?",
+          reply: "Down 1.2kg over three weeks. Keep the weigh-ins to one morning a week.",
+          mutations: [], stateRead: {}, version: i < 12 ? "0683588" : "1c25ce3",
+          lifecycleStatus: null, failureCategory: null, fixRef: null,
+        }));
+        g.__KAMLIFE_STUB_ROWS = new Map<any, any[]>([[schema.turnLedger, clean]]);
+        g.__KAMLIFE_STUB_WRITES = [];
+        await runCoachHealthSweep(1);
+
+        // ONE EVALUATION. The packet must be a by-product of the scan that already ran, and the
+        // audit row is how that is measured — the same signal P0 #115 established.
+        const audits = (g.__KAMLIFE_STUB_WRITES || []).filter((w: any) => w.table === schema.adminEvents);
+        if (audits.length !== 1) {
+          failures.push(`Building the review packet took ${audits.length} audited ledger evaluations — the packet must come out of the scan the sweep already does, not a second one`);
+        }
+
+        const packet = readPacket();
+        if (!packet?.unflagged || !Array.isArray(packet.unflagged.sample)) {
+          failures.push(`The automatic sweep stored no unflagged sample — a new failure shape still has no evidence waiting and the founder is still the transport layer`);
+        } else {
+          if (packet.unflagged.sample.length === 0) {
+            failures.push(`The stored packet's unflagged sample is empty on a window of 24 clean turns — there was evidence to keep and none was kept`);
+          }
+          if (packet.unflagged.clean !== 24) {
+            failures.push(`The packet reports ${packet.unflagged.clean} clean turns out of 24 — the denominator is what tells a reviewer whether 12 turns is a spot check or the whole window`);
+          }
+          const ex = packet.unflagged.sample[0];
+          if (!ex?.turnId || !ex?.version || !ex?.input) {
+            failures.push(`An unflagged sample entry carries no turn/build provenance or no input — it cannot be traced back to the turn it came from`);
+          }
+          // SPREAD, NOT THE HEAD. Ordered newest-first, so a naive slice would sample only the most
+          // recent hours and a failure at the far end of the window could never be in the packet.
+          // With 24 clean turns and a 12-turn sample the stride is 2, so the back half must appear.
+          const ids: string[] = packet.unflagged.sample.map((s: any) => String(s.turnId));
+          if (!ids.some(id => Number(id.slice(-2)) >= 12)) {
+            failures.push(`Every sampled turn came from the newest half of the window (${ids.join(", ")}) — the sample is the head of the list, so the older end of the day is invisible`);
+          }
+          // Deterministic: the same rows must produce the same sample, or two packets cannot be
+          // compared and no sampled turn can be re-derived from the ledger.
+          await runCoachHealthSweep(1);
+          const again: string[] = (readPacket()?.unflagged?.sample || []).map((s: any) => String(s.turnId));
+          if (again.join(",") !== ids.join(",")) {
+            failures.push(`The same 24 rows produced a different unflagged sample on the second sweep — the packet is not reproducible, so a reviewer comparing two runs is comparing dice rolls`);
+          }
+        }
+
+        // BUILD PROVENANCE. buildWarning already said "more than one build served this window"
+        // without ever saying WHICH, and a candidate is only attributable if you know what code
+        // answered the turn.
+        const builds: any[] = packet?.builds || [];
+        if (builds.length !== 2 || !builds.some(b => b.version === "0683588") || !builds.some(b => b.version === "1c25ce3")) {
+          failures.push(`The stored packet does not name the builds that served the window: ${JSON.stringify(builds)} — "a regression may be a turn that ran the old code" is unactionable without them`);
+        }
+
+        // CONTROL: a sample of turns "nothing flagged" must mean it. With every turn flagged there
+        // is nothing clean to keep, and a packet that still produced twelve entries would be
+        // sampling the window rather than the clean part of it.
+        freshTurn();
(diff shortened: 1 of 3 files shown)


ButiMM:
## CTO HOLD — REVIEW PACKET ACCEPTED, #138 RECURRENCE PROOF IS TOO NARROW

The review-packet half is on the right owner and should stay: same sweep rows, no second scan/table/endpoint, durable `scheduler_state` snapshot, build provenance, deterministic unflagged sample, existing audited read surface. Do not unwind that work.

One release-gate requirement is not yet proved.

### The detector currently grades an easier shape than the actual held-constraint defect
`closed-day-food-offer` only fires when `foodDayIsClosed(t.input)` is true on the **same ledger row** whose reply asks for food.

But #138 was a held-constraint failure: the client had already closed the food day, `readHeldConstraints` carried `foodDayClosed: true`, and a **later** SMART NEXT MEAL turn still offered food. The PR body explicitly says the new invariant *cannot* fire if the client closed the day an hour ago. That is precisely the recurrence class #114 required us to make automatically visible.

So do NOT merge yet.

### Repair this PR only
Keep the packet work intact. Extend the recurrence grading so the sweep can detect:

`same user + same SAST food day: earlier inbound closes food day -> later outbound asks/recommends food today`

Use the rows the sweep has already read and the existing food-day semantics. No second DB scan, no new table/state system, no duplicate regex family, no broad held-constraint rewrite. If chronological per-user/day evaluation is the smallest honest shape, do that inside the existing brief/sweep owner.

### Required controls
1. **Actual recurrence:** row A `I'm done eating for today`; later row B `what should I eat?` -> outbound meal/food ask => candidate.
2. **Fixed path:** same earlier closure; later reply stands down / reports totals / points a real gap to tomorrow => clean.
3. **Open-day control:** no prior closure + legitimate meal suggestion => clean.
4. **User/day isolation:** one user's closure must not contaminate another user or a different SAST day.
5. Packet still costs one audited evaluation; unflagged sampling/build provenance/A1-A2 dedup remain green.

Do not widen into the known separate "client changes their mind and reopens food" P2 in this PR. This detector should mirror today's actual held-constraint contract, not solve that future product change.

Also sync/prove against current `main@08228aa` (PR metadata now shows that base, good). Authoritative CI on the current head is still running; after this repair, rerun it and report. One bounded repair, then STOP.

ButiMM:
### Addendum — keep the file governor intact
`server/routes/admin-turns.ts` is already 1196/1200 on this PR. The repair must not layer a second cross-turn detector beside the same-turn one and push the file over budget. Replace the too-narrow recurrence grading with the correct per-user/same-SAST-day shape, or reuse an existing rightful helper if needed. No budget raise, no comment/whitespace compression as the mechanism.

ButiMM:
## CTO hold addressed — the constraint is now graded across turns

Rebased onto `main@08228aa`. Head is now `104c2c7` (`556e732` packet, unchanged + `104c2c7` repair).

The hold was right. My invariant read `foodDayIsClosed` on the **same ledger row** as the food offer — the comment said "ONE TURN, DELIBERATELY" and admitted it could not see a client who closed their day an hour earlier. #138 is precisely that:

```
19:55  "I'm done eating today"      -> the day is closed
20:10  "what should I eat?"         -> nothing here mentions food being closed
20:10  a meal suggestion            -> the violation
```

The offending row's own input says nothing about food, so the narrow detector would have reported that window **clean**. It graded the easy half of the defect.

## What changed — and what did not

**Untouched:** the packet. Same sweep, no second scan, same durable `scheduler_state` snapshot, build provenance, deterministic unflagged sampling, same audited read surface. Nothing rebuilt.

**Changed:** how the constraint is read.

1. **`server/held-constraints.ts` gains `foodCloseLookup(turns)`** — the same rule the rest of the product uses (`foodDayIsClosed`, today only), folded over rows the caller already holds. `readHeldConstraints` answers this for one client and reaches the chat log to do it; Coach Health asks it about thousands of turns it is *already holding*, and a per-turn reader would be thousands of queries for evidence in memory. It belongs in the module named for held constraints, not in the dashboard.
2. **The `Invariant` context gains `at` and `heldFoodClose`** — the one cross-turn fact an invariant may consult. The other five invariants ignore both.
3. **`closed-day-food-offer`** now flags a reply that asks for food when this **client** closed their food day earlier on the **same SAST day**. The old same-row case is subsumed: closure and offer in one turn have equal timestamps, and the guard is `>=`.

**Still no second DB read.** The fold is one in-memory pass over rows the brief had already read. Graded by the audit row, which stays at exactly one per evaluation.

## Controls — each removes a mechanism, and the suite goes red on it

| control | what it removes | failure the suite reports |
|---|---|---|
| **1** | the **old narrow one-turn invariant** | *did not catch the #138 recurrence: the client closed their food day in an earlier turn … and the window was reported clean* |
| 2 | client dropped from the key | *a closed food day for one client flagged **another** client's meal suggestion* |
| 3 | SAST day dropped from the key | *yesterday's closed food day flagged today's meal suggestion* |
| 4 | ordering guard dropped | *a meal suggested **before** the client closed their food day was flagged* |

Control 1 is your objection, executable — the pre-hold code fails the new test with that exact message.

**Healthy, and staying clean:** the reply the merged #138 fix actually produces, on a closed day; a legitimate meal suggestion on an open day.

Graded through `buildCoachHealthBrief` on a seeded multi-turn ledger rather than by calling `holds()` with a hand-made context — the question at issue is whether the sweep assembles the cross-turn fact *itself*, so a fixture that supplied it would be testing the fixture. Turn times are anchored to `sastDayStart`, so no case can pass or fail on the wall clock.

## `admin-turns.ts` held at 1199 / 1200

Moving the fold to its rightful owner is what paid for the new grading; the rest came out of prose **I** wrote in #144, not out of anything the codebase had already earned. The file still wants its own governor cut — it has one line of headroom.

## Verification vs `main@08228aa`

`tsc` clean · tracking-contract **green** · unit **1053/1053** · production-parity **142/142**

| guard | main `08228aa` | branch `104c2c7` |
|---|---|---|
| file-size | utils.ts 1386 ✗ (only) | utils.ts 1386 ✗ (only) |
| `messageDeciders` | 32 / 31 ✗ | 32 ✗ |
| `looksLikePredicates` | 21 / 20 ✗ | 21 ✗ |
| `authorshipPoints` | 423 / 420 ✗ | 423 ✗ |

Zero new failures. No budget raised, nothing suppressed. Requirement 7 is unchanged from the PR body — unattended adjudication is still externally blocked (`DATABASE_URL`/`ADMIN_API_KEY` unset, Railway and `api.openai.com` both `403 CONNECT`), and no credential was added.

---

PR #147, opened 2026-09-03
P0-2: a same-day weigh-in is a durable write, and the turn must record it

ButiMM:
#114 P0-2. Base `main@e66d82f`. `utils.ts` untouched.

## Read this first: the defect as adjudicated does not reproduce

A same-day weight correction **is** written. `handleWeightLog` has always updated today's row, and `users.currentWeight` moves with it:

```
84.0 already logged today, client sends "83.4kg"
  weight_logs   UPDATE -> 83.4   (no second row)
  users         currentWeight -> 83.4
```

**My acceptance trace was wrong**, and how it was wrong matters. It read `__KAMLIFE_STUB_WRITES` — inserts only. This path `UPDATE`s, so the correction was invisible to the instrument and I reported "acknowledged but never written" from an absence I had manufactured. Worse: the steps section of `tracking-contract-tests` already carries the warning — *"Both channels must be read, or an update looks exactly like a no-op."* I made that exact mistake one suite over.

The corroborating signal I leaned on, `[TURN_OWED] workout stood down — weight stated and not yet written`, was not independent evidence either. It comes from the same missing record described below.

## What was real, in the same three lines

`turnMutation` sat only in the **insert** branch. So a same-day weigh-in changed the ledger while the turn recorded `mutations: null`, and every reader of that record was working from "this turn wrote nothing":

| reader | consequence |
|---|---|
| `turnAlreadyWrote("weight")` | stayed false — a second door could write the day again |
| routes' `wroteThisTurn` | stayed false — claimants that must stand down after a durable write did not |
| the workout owner | logged *"weight stated and not yet written"* — a decision on a false premise |
| Coach Health | reads `mutations`, so LAW 4 could not see a same-day weigh-in **at all** |

The write and the record of the write are one fact.

## The change

The update branch now calls `turnMutation` like every other writer, worded UPDATE-with-was to match the steps owner so a reader can tell a correction from a first weigh-in:

```
INSERT weight=83.4kg                first weigh-in of the day
UPDATE weight=83.4kg (was 84kg)     a correction
UPDATE weight=84kg                  an unchanged repeat
```

`weight` now comes back beside `id` from the query that was **already running** — one more column, no extra query, and the turn note is its only consumer.

## Controls — in `tracking-contract-tests`, reading both channels

| # | case | asserted |
|---|---|---|
| 1 | first weight of the day | one row added · `INSERT` recorded |
| 2 | changed same-day weight | row corrected · **no** row added · `users.currentWeight` follows · `UPDATE` recorded **with what it replaced** |
| 3 | unchanged repeat | no row added · no `INSERT` recorded |

Plus the implication the adjudication asked for: *if the reply says 83.4 back, 83.4 is what the ledger holds.*

**Red on revert:**

| control | removed | failure |
|---|---|---|
| **A** | `turnMutation` from the update branch | *recorded no durable write on the turn: `[]`* |
| **B** | correcting the row (always INSERT) | *added 1 weight rows — the day must hold one weigh-in, corrected, not two* |

Control B matters because without it "always insert" would satisfy case 2 by writing a duplicate row, leaving the day holding two conflicting weights.

### One thing the offline harness cannot prove

`getWeightTruth` reading the corrected row back is **not** verified here — the stub does not reflect `UPDATE`s into subsequent reads. What is verified is that the update targets today's row with the new value and that `users.currentWeight`, which most downstream surfaces read, carries it. Saying so rather than implying full coverage.

## Verification vs `main@e66d82f`

`tsc` clean · tracking-contract **green** · unit **1053/1053** · production-parity **142/142** · gap-tests **338/338** · routing-audit green

| guard | main | this branch |
|---|---|---|
| file-size | **236 files OK (none over budget)** | 236 files OK |
| `messageDeciders` | 32 / 31 ✗ | 32 ✗ |
| `looksLikePredicates` | 21 / 20 ✗ | 21 ✗ |
| `authorshipPoints` | 423 / 420 ✗ | 423 ✗ |

No budget raised, nothing suppressed. `weight.ts` 513 lines against the 1200 default.

**If you would rather this landed as controls only, without the `turnMutation` line, say so** — the controls stand on their own and lock behaviour that was previously untested either way.

---

### DIFF
--- a/script/tracking-contract-tests.ts
+++ b/script/tracking-contract-tests.ts
@@ -322,6 +322,92 @@ const MUST_WRITE: [string, string][] = [
   }
 
   /**
+   * A SECOND WEIGH-IN TODAY IS A DURABLE WRITE, AND THE TURN MUST SAY SO (#114 P0-2, 2026-09-03).
+   *
+   * The adjudicated report said a same-day weight correction was acknowledged but never written.
+   * It IS written — handleWeightLog has always updated today's row. The report was wrong because
+   * the trace behind it read only INSERTs, and this path UPDATEs. The same mistake the steps
+   * helper above warns about, made one suite over.
+   *
+   * What was real: turnMutation sat only in the insert branch, so the correction changed the
+   * ledger while the turn recorded NOTHING. turnAlreadyWrote("weight") stayed false, routes'
+   * wroteThisTurn stayed false, and Coach Health — which reads `mutations` — could not see a
+   * same-day weigh-in at all. The write and the record of the write are one fact.
+   *
+   * Graded on both channels and on the turn's own record, because either alone lies here.
+   */
+  {
+    const weighIn = async (todayRow: any | null, message: string) => {
+      freshTurn();
+      g.__KAMLIFE_STUB_USER = { ...USER, currentWeight: "86.0", heightCm: 178, gender: "male", age: 35 };
+      const rows = [{ id: "wl-old", userId: USER.id, weight: "86.0", weightKg: "86.0",
+        loggedAt: new Date(dayStart.getTime() - 7 * 86_400_000), at: new Date(dayStart.getTime() - 7 * 86_400_000) }];
+      if (todayRow) rows.push(todayRow);
+      g.__KAMLIFE_STUB_ROWS = new Map([[schema.weightLogs, rows]]);
+      g.__KAMLIFE_STUB_WRITES = [];
+      g.__KAMLIFE_STUB_UPDATES = [];
+      const reply = String(await handleMessage(USER.phoneNumber, message).catch(() => ""));
+      const ins = (g.__KAMLIFE_STUB_WRITES || []).filter((w: any) => w.table === schema.weightLogs);
+      const upd = (g.__KAMLIFE_STUB_UPDATES || []).filter((w: any) => w.table === schema.weightLogs);
+      const usr = (g.__KAMLIFE_STUB_UPDATES || []).filter((w: any) => w.table === schema.users)
+        .map((w: any) => w.set?.currentWeight).filter(Boolean);
+      const muts: string[] = ((g.__KAMLIFE_STUB_WRITES || [])
+        .filter((w: any) => w.table === schema.turnLedger)
+        .flatMap((w: any) => (Array.isArray(w.values?.mutations) ? w.values.mutations : []))) as string[];
+      delete g.__KAMLIFE_STUB_UPDATES;
+      delete g.__KAMLIFE_STUB_ROWS;
+      return {
+        reply, inserted: ins.length, rowsAdded: ins.length,
+        stored: ins.length ? String(ins[0].values?.weight) : upd.length ? String(upd[upd.length - 1].set?.weight) : null,
+        currentWeight: usr.length ? String(usr[usr.length - 1]) : null,
+        weightMutations: muts.filter(m => /weight=/i.test(m)),
+      };
+    };
+    const TODAY_84 = { id: "wl-today", userId: USER.id, weight: "84.0", weightKg: "84.0",
+      loggedAt: new Date(dayStart.getTime() + 60_000), at: new Date(dayStart.getTime() + 60_000) };
+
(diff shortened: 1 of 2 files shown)


ButiMM:
CTO HOLD — one bounded semantic repair. Keep the diagnosis correction and the genuine same-day correction fix.

The real defect is valid: when today's weight changes (84 -> 83.4), canonical state changes and the turn must record a weight UPDATE mutation so turnAlreadyWrote / wroteThisTurn / Coach Health can see the durable write.

But the current implementation also records `UPDATE weight=84kg` when the client simply repeats the exact same 84kg. That contradicts this PR's own control text: "AN EXACT REPEAT IS NOT A SECOND WEIGH-IN" / "unchanged repeat stays deduped." `mutations` is operational state consumed by downstream claimant stand-down and Coach Health; it is not harmless logging.

Repair only this:
1. In the existing-today branch, record a weight `turnMutation` only when the stored numeric value actually changes (or the stored value is unusable and is being repaired). Preserve the existing database behaviour unless skipping an identical SQL update is demonstrably safer; the contract is that a semantic no-op must not become a durable turn mutation.
2. Keep changed-correction wording `UPDATE weight=83.4kg (was 84kg)`.
3. Tighten case 3: exact repeat must assert ZERO weight mutations, not merely zero INSERTs. This is the over-fire control.
4. Re-run the same suites and authoritative CI. No widening.

Do NOT turn this into controls-only: the missing turn mutation on a genuine correction is a real current defect. Do NOT start the next P1 until this PR is repaired/pushed and CI is re-triggered.

ButiMM:
## Repaired — head `738b74d`, CI re-triggered

The hold was right, and the contradiction was in this PR's own text: the control said *"AN EXACT REPEAT IS NOT A SECOND WEIGH-IN"* and the implementation recorded `UPDATE weight=84kg` for exactly that. `mutations` is operational state — claimant stand-down and Coach Health both consume it — so a semantic no-op entering it stands doors down for a turn that changed nothing and hands Coach Health a weigh-in that never happened.

### 1 · Recorded only when the day actually moved

```
84.0 -> 83.4          UPDATE weight=83.4kg (was 84kg)     recorded
84.0 -> 84.0          nothing                             not a write
unparseable -> 84.0   UPDATE weight=84kg                  recorded, no "was"
```

The unparseable case is the exception you allowed: that update **is** a repair and does change the day, so it is recorded — without a `was` it cannot honestly state.

**SQL left alone.** The row is still set to the value it already held. Narrowing the write is a separate question from narrowing the record, and this PR only claims the second.

### 2 · Changed-correction wording kept

`UPDATE weight=83.4kg (was 84kg)` — unchanged, and still asserted.

### 3 · Case 3 tightened to the over-fire control

Now asserts **zero weight mutations**, not zero INSERTs. The weaker form passed the defect you caught, which is the whole reason it needed raising.

The rule is now pinned from both sides:

| control | failure |
|---|---|
| record on **every** same-day write | *Repeating today's weight recorded `["UPDATE weight=84kg (was 84kg)"]` as a durable write — nothing about the day changed, and claimant stand-down and Coach Health both read that record* |
| record on **none** | *A same-day weight correction recorded no durable write on the turn: `[]`* |

### 4 · Suites re-run vs `main@e66d82f`

`tsc` clean · tracking-contract **green** · unit **1053/1053** · production-parity **142/142** · gap-tests **338/338** · routing-audit **318/318**

| guard | main | branch |
|---|---|---|
| `messageDeciders` · `looksLikePredicates` · `authorshipPoints` | 32 ✗ · 21 ✗ · 423 ✗ | identical |

No widening, no budget raised, `utils.ts` untouched. Not starting the next P1 until this is merged.

### Ledger correction, recorded here as well as in the PR body

The alleged P0 **"same-day weight correction is acknowledged but not stored" is false** and should come off the defect ledger. The real defect is: **the same-day correction persisted correctly, but its durable-write evidence was missing from the turn.** My original trace read only `__KAMLIFE_STUB_WRITES` (inserts) on a path that `UPDATE`s, and I reported an absence I had created.

---

PR #150, opened 2026-09-03
Expose effective runtime modes on health

ButiMM:
Implements only the `/health` runtime-provenance cut from #114 on `main@4a4a510`.

- adds the effective Meaning Engine and normalizer modes to the existing build snapshot returned on every `/health` path
- reads `ENGINE_LIVE` through its existing `engineLive()` owner
- gives the existing normalizer owner one reusable live-mode helper, preserving the current exact `NORMALIZER=off` killswitch semantics
- keeps the existing endpoint, build SHA, startup state, database behavior, and product policy unchanged
- adds runtime controls for engine-on/normalizer-off and engine-off/normalizer-default-on

Verification before push:
- `npm run check` - PASS
- `npm run check:names` - PASS (97 ? 97)
- unit merge-base comparator - base 1051/1055, head 1051/1055, NEW 0
- focused gap suite - new control PASS; existing unrelated Windows hunger child-process failure remains

No architecture consolidation and no product/health-policy change are included.

Refs #114

PR #151, opened 2026-09-03
P1: a follow-up that names no subject belongs to the conversation, not to a topic

ButiMM:
#114 P1. Base `main@4a4a510`. 3 files.

## What the trace found

The founder asked what the coach's answer meant, right after being told *"7.0kg to go: 92kg now, 85kg the goal"*, and lost the weight conversation. I traced the **claimant** rather than the reply text, and there were two owners with one cause:

| # | owner | what it did |
|---|---|---|
| 1 | `numbers-literacy` | claimed `what does that mean` as a **calorie** question and answered *"you never have to understand calories or count anything"* |
| 2 | `early-commands` | then read it as **confusion** and answered with the day's action |

**1 — the alternation was `what does (that|this|the number|kcal|calories?) mean`.** So a bare pronoun claimed the turn with no calorie word anywhere in it. `the number`, `kcal` and `calories` say what they are about and still claim; `that` and `this` say nothing. The second clause is untouched and still catches a pronoun beside a real calorie word.

**2 — that branch already carries the rule**, written after a live 2026-07-31 failure: *"A QUESTION IS NEVER CONFUSION … never answer a real question with the day's action."* It tested it with `/\?/.test(message)` — which is the **punctuation line of `utils.looksLikeQuestion` without its interrogative openers**. So the same question asked without a `?` was read as confusion. The canonical owner decides it now; a second spelling of "is this a question" inside a handler is what let it through.

```
before   "what does that mean"  ->  CALORIE_EXPLAINER, then CONFUSED_ONE_ACTION
after    "what does that mean"  ->  the coach, with the client's snapshot and
                                    memory — exactly like "what does it mean?"
```

Both spellings of the follow-up now reach the same place, which is the coherence this was raised for.

## Controls — in `routing-audit`, graded on each claimant's own copy

| control | asserts |
|---|---|
| **reject** | `what does that mean` and `what does this mean?` get neither the calorie lecture nor the confusion lead |
| **over-fire** | `what does the number mean?` still reaches the explainer, and so does `I am confused about calories, what does that mean` |
| **red on revert** | pronoun back in the calorie alternation → **both** pronoun cases fail; `/\?/` back in the confusion guard → **only** the no-question-mark case fails |

The over-fire pair matters because without it "never claim" would pass the two rejects. The two red-on-revert controls redden *different* cases, which is what proves each half is separately load-bearing rather than one masking the other.

## What this PR does not claim

Offline the model is unreachable, so a turn that now reaches the coach degrades to the deterministic fallback. **That the answer then speaks to the prior subject is not asserted here and cannot be from this environment** — the recorded egress blocker. What *is* asserted is where the turn goes, which is the divergence and is fully deterministic.

One related observation I did **not** act on, to keep this bounded: the punctuation-only follow-up (`???`) gets an explicit *"they are responding to your previous message"* context built from the last exchange, and the worded follow-up does not. That is a context-assembly difference inside the GPT block, a separate owner and a separate cut.

## Verification vs `main@4a4a510`

`tsc` clean · **routing-audit 322/322** (318 + 4) · unit **1055/1055** · production-parity **142/142** · tracking-contract green · gap-tests **344/344** · food-scanner **48/48**

| guard | main | branch |
|---|---|---|
| `messageDeciders` · `looksLikePredicates` · `authorshipPoints` | 32 ✗ · 21 ✗ · 423 ✗ | identical |

No budget raised, no new predicate, no new module — the fix uses `utils.looksLikeQuestion`, which already existed, and removes a partial duplicate of it.

---

### DIFF
--- a/script/routing-audit.ts
+++ b/script/routing-audit.ts
@@ -598,8 +598,27 @@ const CASES: Case[] = [
     const SESSION_SERVE = /Reply \*DONE\* when finished|Next Session \(Day/i;
     const SESSION_LOGGED = /Session \d+ (?:logged|—|in)\b|First workout done|got it, logged to/i;
     const FOOD_LOGGED = /\*Food logged|Food logged ✅|Running total today|Meal total:/i;
+    const CALORIE_LECTURE = /you never have to understand calories or count anything/i;
+    const CONFUSION_LEAD = /here'?s the one that matters/i;
     const x = (name: string, msg: string, c: Partial<Case> = {}): Case => ({ name: `xintent: ${name}`, msg, ...c });
     return [
+      // A BARE PRONOUN NAMES NO SUBJECT (#114 P1, 2026-09-03, founder). "what does that mean"
+      // was claimed by the calorie explainer with no calorie word in it, so a client who had just
+      // been told "7.0kg to go: 92kg now, 85kg the goal" and asked what that meant was answered
+      // "you never have to understand calories or count anything" — a lecture on a subject nobody
+      // had raised. Without a "?" it was then read as CONFUSION and answered with the day's action.
+      // Two owners, one cause: a follow-up whose subject is whatever was just discussed belongs to
+      // the path that can see the conversation, not to a topic handler or to the daily ladder.
+      x("bare pronoun follow-up is not a calorie question", "what does that mean",
+        { reject: [CALORIE_LECTURE, CONFUSION_LEAD] }),
+      x("bare pronoun follow-up, with a question mark", "what does this mean?",
+        { reject: [CALORIE_LECTURE, CONFUSION_LEAD] }),
+      // OVER-FIRE CONTROLS. The explainer still owns a question that says what it is about —
+      // without these, "never claim" would pass the two cases above.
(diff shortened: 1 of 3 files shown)


ButiMM:
CTO HOLD / SYNC ONLY — do not start another product cut yet.

PR #150 has now merged to main as `401bc9d`, so this PR is no longer mergeable on its old base `4a4a510`.

Do only this:
1. fetch/sync current `origin/main@401bc9d`;
2. rebase/update this existing #151 branch onto that main with no widening of the claimant fix;
3. resolve only genuine merge conflicts if any, preserving #150's `/health` runtime provenance and this PR's two claimant corrections;
4. rerun TypeScript, routing-audit, unit-baseline and authoritative CI;
5. push this same PR and STOP.

Do NOT add the separate punctuation/context-assembly observation yet. Do NOT start food-day reversal. The remaining adjudication depends on production `/health.runtime`: we need to know whether `ENGINE_LIVE` is actually on before deciding whether this claimant repair fully closes the subject-continuity P1 or needs one final deterministic context seam.

Issue #164, opened 2026-09-04
#158 final proof: targeted multi-day food correction on real PostgreSQL


ButiMM:
## CTO CONTINUATION — CLAUDE LANE, START NOW

PR #161 has merged to `main@70f61a3`. Continue the SAME recovered #158 Event Model/correction contract. Do not start architecture or normalizer work; Codex owns those lanes.

The remaining unproved contract is the one Claude correctly refused to fake in #159/#161:

1. client sends a multi-day catch-up;
2. durable rows now preserve per-day structured `items` after #161;
3. client later says `Tuesday wasn't rice, it was pap.`;
4. the correction must target ONLY Tuesday's intended row/event under REAL PostgreSQL ordering/update semantics.

The current deterministic fixture cannot prove this because it ignores `orderBy` and does not faithfully reflect UPDATE reads. That limitation is now the task, not an excuse to infer correctness.

### 1. Use real PostgreSQL, not a smarter fake

First inspect whether the repository already has an existing real-DB test owner/harness. Reuse it if it exists.

If there is no existing owner, add the smallest **ephemeral PostgreSQL acceptance path** needed for this exact contract. Preferred shape: a PostgreSQL service/container in CI feeding the real Drizzle/database layer and existing migrations/schema, with one focused current-pipeline test. This is real PostgreSQL semantics, not a second in-memory database implementation.

Do NOT build a general new testing framework. Do NOT modify the ordinary deterministic DB stub to emulate `ORDER BY`/UPDATE behavior.

### 2. Seed through the product path

On current main, drive the actual `handleMessage`/food pipeline for one user:

`Monday I had eggs and toast. Tuesday I had rice and chicken. Wednesday I had pap and livers.`

Read the actual PostgreSQL rows after the turn and prove:
- exactly the intended three historical food rows exist;
- three distinct resolved dates;
- one shared source/event group where current semantics require it;
- every row has the structured items now delivered by #161;
- Tuesday contains rice/chicken before correction;
- Monday and Wednesday have distinct, preserved identities.

Snapshot the full relevant state of all three rows before correction: IDs, dates, labels, sourceMessageId, totals, items, corrected flag/metadata and any fields the correction owner can mutate.

### 3. Run the named correction through the real path

Send exactly:

`Tuesday wasn't rice, it was pap.`

Then prove from PostgreSQL, not logs alone:
- no duplicate Tuesday row was inserted;
- exactly the intended existing Tuesday row was updated;
[...]

Issue #158, opened 2026-09-04
Recovered contract replay: food events and corrections must preserve identity, day and one final reply


ButiMM:
## CTO execution order — current main only

This is the next recovered contract from historical PR #48. **Do not merge, rebase, cherry-pick, or copy the old branch.** It is hundreds of commits behind current main and its implementation is not authoritative.

Start from current `origin/main@4610be1` (or newer if main has advanced before you begin).

The job is **REPLAY FIRST**. Prove what current main does. If every required contract already holds, make no product change: add only the smallest missing regression evidence if genuinely absent, report GREEN, and stop. If one fails, fix the **first current-main divergence only**, in the existing food/event/correction owners, with one bounded PR and red-on-revert proof.

## Historical contract being recovered

Old #48 was about a real product boundary: one client message can contain multiple food events, corrections, dates and non-food clauses. The stored truth must preserve the identity of each event so a later correction can edit the right thing without corrupting another meal/day, and the client still receives one coherent final reply.

A historical unresolved risk was multi-day catch-up rows with `items: []`: the correct day could be found later, but there was nothing structured to edit when the client named a food correction. **Do not assume that defect still exists. Reproduce it on current main.**

## Required current-main replay cases

Use the real current pipeline and the strongest existing harness available (real PostgreSQL where that is already the owner; otherwise the current deterministic fixture only if it can honestly represent the state). Read both INSERT and UPDATE channels where relevant.

1. **Two eating events in one message**
   - Example shape: `Breakfast was 2 eggs and toast. Lunch was pap and chicken.`
   - Must persist as the correct distinct eating events/slots or equivalent current event model truth; no double charge, no event collapse that prevents later correction.
   - Client-facing turn closes once: no duplicate receipt/reply mouths.

2. **Intra-message correction**
   - Example: `Breakfast was 2 eggs — actually 3 eggs. Lunch was pap and chicken.`
   - Net truth must contain the corrected breakfast, not both stale and corrected versions.
   - Lunch must remain untouched.

3. **Food + non-food clause isolation**
   - Example: `Breakfast was 2 eggs and toast, and I did 5000 steps.`
   - The step clause must not become a food item/event; the food event must not swallow/drop the step fact.
   - Preserve the current multi-intent one-final-reply contract.

4. **Multi-day catch-up preserves structured identity**
   - Example shape: `Monday I had eggs and toast. Tuesday I had rice and chicken. Wednesday I had pap and livers.`
[...]

PR #165, opened 2026-09-04
Close release architecture governor honestly

ButiMM:
Closes #162.

Base: origin/main@70f61a389536b09436f03277ade51ff7122441d1

## Structural adjudication

This changes only script/check-architecture.ts. No product code, routing, predicates, copy, or tests changed.

- messageDeciders: measured 32. The last legitimate 31-file baseline and current inventory differ only by server/handlers/chat-log.ts. Its meaningful-turn audit decision and stored-media-receipt decision are both live, distinct, and owned at the final-turn/history boundary. Consolidation candidates were traced and rejected because they answer different questions or would move the decision to the wrong owner. Minimal truthful rebaseline: 31 ? 32, with a ratchet entry.
- looksLikePredicates: measured 21. Since the original 20 freeze, looksLikeSameMeal was removed and two distinct live boundaries were added: bulk first-contact intake and grounded recall. Both have production callers and cannot be combined without removing behavior or violating the grounded-recall boundary. Minimal truthful rebaseline: 20 ? 21, with a ratchet entry.
- authorshipPoints: measured 423. Three exact matches are machine values, not client-authored prose: isConnectionError returns boolean membership, holdTokens returns persisted profileNotes storage syntax, and cacheKey returns a private dedupe key. The matcher now excludes only those exact returned lines. The budget remains frozen at 420; any other return in those files still counts.

No behavior was deleted, no symbol was renamed, no new module or predicate was added, and no red was traded into another governor.

## Exact message-decider inventory - 32

server/condition-welcome.ts
server/exercise-variants.ts
server/food-referent.ts
server/food-swaps.ts
server/gpt.ts
server/handlers/advice-commands.ts
server/handlers/chat-log.ts
server/handlers/conversion.ts
server/handlers/early-commands.ts
server/handlers/food-commands.ts
server/handlers/food-context.ts
server/handlers/food-log-mgmt.ts
server/handlers/gpt-block.ts
server/handlers/lifecycle.ts
server/handlers/meal-repeat.ts
server/handlers/misc-commands.ts
server/handlers/numbers-literacy.ts
server/handlers/pain-triage.ts
server/handlers/reminders-handler.ts
server/handlers/safety.ts
server/handlers/sick-flow.ts
server/handlers/water.ts
server/handlers/workout.ts
server/medication-context.ts
server/memory.ts
server/numbers-mode.ts
server/reminders-parse.ts
server/routes.ts
server/safety-detection.ts
server/understanding/actions.ts
server/utils.ts
[...]

PR #166, opened 2026-09-04
#164: a named-day food correction searches the day they named — proved on real PostgreSQL

ButiMM:
Issue #164 — the final #158 proof. Base `main@70f61a3`, head **`aefeb2e`** (verified with `git rev-parse`). 3 files.

**PostgreSQL 16.13** (`postgres:16` in CI; locally the same version), project schema applied with `drizzle-kit push`, real `handleMessage` pipeline, rows read back with SQL. The database is authoritative — not the logs.

## First divergence

`applyIdentityCorrection` selected candidates with `loggedAt >= sastDayStart()` — **today, always**. A correction naming a past day found nothing, returned `null`, and fell through to the food scanner:

```
before   Mon 360/24 [Eggs on toast]     grp=a14ff9
         Tue 580/56 [Chicken and rice]  grp=a14ff9
         Wed 593/51 [Pap, Ox liver]     grp=a14ff9

"Tuesday wasn't rice, it was pap."

after    Tue 580/56 [Chicken and rice]  grp=a14ff9   <- untouched
         Tue 330/7  [Pap] label=snack   grp=f7c1fc   <- a SECOND row
         reply: "Got it — Pap and Rice … could not price *tuesday, wasn*"
```

Four rows, the denied rice still there, a new event group, and the client told their correction was a new meal.

## The fix, in that owner

The window is the day they named. `parseMealDate` is the day owner this file already imports and the multi-day writer already uses, so the named day resolves exactly as it was written. A correction naming no earlier day keeps the window it had — today, unbounded above — so **every same-day correction is byte-identical to before**.

```
after    Mon 360/24 [Eggs on toast]  grp=02a302  corrected=false
         Tue 578/12 [Pap]            grp=02a302  corrected=true   <- same row id
         Wed 593/51 [Pap, Ox liver]  grp=02a302  corrected=false
         reply: "Fixed — logged as *Pap*, not Chicken and rice."
```

## Ten checks, all read from SQL

three rows before and after, no duplicate · the **same** Tuesday id updated, not replaced · denied rice gone, pap present · Monday byte-identical to its pre-correction snapshot · Wednesday byte-identical · today neither created nor mutated · event lineage preserved · one final client reply · and #159's over-fire contract: `it wasn't that bad honestly` mutates nothing.

**Red on revert:** restoring the today-only window fails four of them, including the duplicate row and the surviving rice.

## A second divergence — found by my own isolation control, not fixed here

`"Monday wasn't toast, it was rice."` never reaches this owner. A different one claims it first:

```
[MEAL_CORRECT] removed=[] added=[rice] day=Wed Sep 02→Wednesday 593→813
```

It targets the **newest** row rather than the named day, and **adds** the new food instead of replacing the denied one. Monday keeps its toast; Wednesday gains rice.

#164 scopes to the first divergence, so the acceptance script **reports** this rather than asserting it — a red there would hide the contract that *is* proved, and deleting it would lose the finding. It is the next divergence, with its reproduction preserved in the script.

## CI

One scoped job: an ephemeral `postgres:16` service, `drizzle-kit push`, one focused script. Every pure suite still runs with no database and no network; `unit-baseline` and the ordinary test/Gauntlet path are untouched. **No secret is required** — the credentials are the throwaway ones the container is created with, and no production or Railway data is reachable from it. The script **skips loudly** without `DATABASE_URL` rather than passing on nothing.

## Not adjusted, and outside this cut

The correction preserves the row's **calories** when substituting, so Tuesday's protein moved 56 → 12 with the food; and the confirmation still closes with *"Today: 0 kcal"* while correcting a past day. Both are pre-existing behaviour of the correction owner.

## Verification vs `main@70f61a3`

`tsc` clean · **pg-correction-acceptance GREEN** on PostgreSQL 16.13 · unit **1057/1057** · production-parity **142/142** · routing-audit **322/322** · gap-tests **350/350** · food-scanner **48/48** · tracking-contract green

| guard | main | branch |
|---|---|---|
| `messageDeciders` · `looksLikePredicates` · `authorshipPoints` | 32 ✗ · 21 ✗ · 423 ✗ | identical — untouched |
| ownership · names · reach · SAST | OK · 97/97 · 8 · 61/61 | identical |

No budget raised, no architecture or normalizer work, no new correction engine/parser/store, no fake `ORDER BY` layer in the stub.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,3 +79,39 @@ jobs:
         if: ${{ !cancelled() && steps.dependencies.outcome == 'success' }}
         run: npm run gauntlet
         continue-on-error: true
+
+  # ── REAL POSTGRESQL, FOR THE ONE CONTRACT THE FIXTURE CANNOT PROVE (#164) ──────────────────
+  #
+  # The deterministic stub ignores ORDER BY and does not reflect UPDATEs into reads, so it cannot
+  # say WHICH row a named-day food correction lands on. Two PRs correctly refused to infer that
+  # (#159, #161); this job is the answer rather than a smarter fake.
+  #
+  # SCOPED ON PURPOSE. One ephemeral service container, one focused script, its own job — every
+  # pure suite above still runs with no database and no network. Nothing here needs a secret:
+  # the credentials are the throwaway ones this container is created with, and no production or
+  # Railway data is reachable from it.
+  pg-acceptance:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: kam
+          POSTGRES_PASSWORD: kam
+          POSTGRES_DB: kamlife
+        ports: ["5432:5432"]
+        options: >-
+          --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      DATABASE_URL: postgres://kam:kam@127.0.0.1:5432/kamlife
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+      - run: npm ci
+      - name: Apply the project schema to the ephemeral database
+        run: npx drizzle-kit push --force
+      - name: Targeted multi-day food correction on real PostgreSQL
+        run: npx tsx script/pg-correction-acceptance.ts
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -111,7 +111,11 @@ jobs:
           node-version: 20
           cache: npm
       - run: npm ci
-      - name: Apply the project schema to the ephemeral database
-        run: npx drizzle-kit push --force
+      # COMMITTED MIGRATIONS, THE SAME ONES A DEPLOY APPLIES. check-schema-safety forbids the
+      # schema-sync shortcut in any workflow, for the reason it names: a bad sync corrupts the
+      # day-ledger and the only way back is the 6-hourly backup. It also scans this file as text,
+      # so the forbidden command is not written here even in a comment.
+      - name: Apply committed migrations to the ephemeral database
+        run: npm run db:migrate
       - name: Targeted multi-day food correction on real PostgreSQL
         run: npx tsx script/pg-correction-acceptance.ts
--- a/migrations/meta/_journal.json
+++ b/migrations/meta/_journal.json
@@ -43,6 +43,20 @@
       "when": 1787100000000,
       "tag": "0005_adaptive_baseline",
       "breakpoints": true
+    },
+    {
+      "idx": 6,
+      "version": "7",
+      "when": 1787200000000,
+      "tag": "0006_durable_client_facts",
+      "breakpoints": true
+    },
+    {
+      "idx": 7,
+      "version": "7",
+      "when": 1787300000000,
+      "tag": "0007_turn_triage",
+      "breakpoints": true
     }
   ]
 }
--- a/script/pg-correction-acceptance.ts
+++ b/script/pg-correction-acceptance.ts
@@ -93,31 +93,8 @@ await say("Monday wasn't toast, it was rice.");
 const afterAdj = await rows();
 const monAfter = day(afterAdj, "Mon")[0], wedAfter = day(afterAdj, "Wed")[0];
 chk(afterAdj.length === 3, `adjacent-day correction added no row (${afterAdj.length})`);
-/**
- * KNOWN, NOT FIXED HERE — a SECOND divergence, in a different owner.
- *
- * #164 asked for the first divergence only, and that one (the named-day search window in
- * applyIdentityCorrection) is fixed and proved above. This case is reported rather than asserted
- * so it cannot be lost, and so a red here does not hide the contract that IS proved.
- *
- * "Monday wasn't toast, it was rice." does not reach applyIdentityCorrection at all: a different
- * owner claims it first and logs
- *
- *     [MEAL_CORRECT] … removed=[] added=[rice] day=Wed Sep 02→Wednesday kcal=593→813
- *
- * so the correction lands on the NEWEST row rather than the day the client named, and it ADDS the
- * new food instead of replacing the denied one. Monday keeps its toast; Wednesday gains rice.
- */
-const monMoved = !!monAfter && monAfter.id === monBefore.id && monAfter.corrected === true;
-const wedIntact = snap(wedAfter) === wedBefore;
-if (!monMoved || !wedIntact) {
-  REAL(`\n  ⚠ KNOWN, NOT FIXED HERE — a named-day correction on an ADJACENT day is claimed by a`);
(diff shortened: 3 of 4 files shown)


ButiMM:
CTO HOLD — product fix and real-PostgreSQL targeting proof are accepted in shape, but authoritative CI found one separate release-safety blocker in the new test infrastructure.

`check-schema-safety` is RED because `.github/workflows/test.yml` runs `npx drizzle-kit push --force`. This repo explicitly forbids `drizzle-kit push` in CI/deploy workflows; committed migrations are the only approved schema path. The guard's reason is production data safety, and we do not bypass it even for an ephemeral acceptance job.

Repair ONLY the CI schema setup:
1. replace the `drizzle-kit push --force` step with the repository's approved committed-migration path (`npm run db:migrate`, or the exact existing migration command if the script name resolves differently on current main);
2. keep the ephemeral `postgres:16` service and focused `pg-correction-acceptance.ts` job otherwise unchanged;
3. prove the committed migrations bring a blank PostgreSQL 16 database to the schema required by this acceptance path;
4. rerun the real PostgreSQL acceptance, schema-safety, ordinary test job, unit-baseline and Gauntlet;
5. push this SAME PR and STOP.

Do NOT touch the named-day product fix, do NOT fix the newly-discovered Monday/newest-row second divergence in this PR, and do NOT weaken/exclude the schema-safety guard. That second correction-owner defect is preserved as the next Claude job after #166 merges.

Current authoritative evidence on `aefeb2e`: pg-acceptance SUCCESS; unit-baseline SUCCESS; TypeScript SUCCESS; all functional suites through production-parity green; Gauntlet executed; test RED on (a) this new schema-safety violation and (b) the standing architecture governor only. Normalizer replay remains non-vacuous debt/covered nothing.

ButiMM:
CTO RE-ADJUDICATION — the prior HOLD is expanded by the automated review finding. #166 is NOT done yet.

Two blockers must be repaired on this SAME PR before merge:

1. **Schema-safety:** replace `npx drizzle-kit push --force` in the PostgreSQL CI job with the repo-approved committed migration path (`npm run db:migrate` / exact existing migration command). Do not weaken the guard.

2. **The PostgreSQL acceptance must bite on the adjacent-day corruption it already observes.** The review is correct: `Monday wasn't toast, it was rice.` currently leaves Monday unchanged and adds rice to Wednesday, but the script only prints a warning, so `pg-acceptance` exits GREEN while the ledger is wrong. That violates #164's required adjacent-day isolation control and the repository's rule that a control which cannot fail is worthless.

Repair the existing composition-correction owner in `server/handlers/food-log-mgmt.ts` so an explicit named day constrains candidate row selection there too. The current branch selects the globally newest row with `.orderBy(desc(loggedAt)).limit(1)` and only respects `parseMealDate(m)` when `plan.moves` is true; that is why Monday mutates Wednesday. Use the existing day owner and correction machinery. No new parser/engine/store/state machine.

Then turn `monMoved` and `wedIntact` into BLOCKING assertions. Required PostgreSQL proof:
- Tuesday identity correction still updates the same Tuesday row only;
- `Monday wasn't toast, it was rice.` updates the same Monday row, removes toast, adds rice;
- Wednesday is byte/field-identical to its pre-correction snapshot;
- no duplicate row, no today row;
- same-day/no-explicit-day correction unchanged;
- non-correction `wasn't` sentence mutates nothing;
- one final reply;
- red-on-revert of newest-row selection fails the Monday isolation case.

Also sync/rebase onto current `main@8c8cce1` (architecture #165 is now merged), so the rerun must inherit the newly GREEN architecture gate.

Do NOT split the Monday defect into a later PR now; the real-DB acceptance already contains it and #164 explicitly required this isolation. Fix both blockers, rerun full CI + pg-acceptance + unit-baseline + Gauntlet, push #166, and STOP.

ButiMM:
## Both blockers repaired — rebased onto `main@8c8cce1`, head `df2b0c5` (verified)

### 1 · Committed migrations in CI — and what that exposed

The job now runs `npm run db:migrate`. `check-schema-safety` **did** fail on the old line, and on my first replacement too: the guard scans the workflow as text, so even a comment naming the forbidden command trips it. Reworded.

**Switching exposed a real defect, and it is the more serious half of this commit.** A database built from committed migrations could not accept a user row at all:

```
column "dietary_restrictions" of relation "users" does not exist
```

`migrations/meta/_journal.json` listed 0000–0005. **`0006_durable_client_facts.sql` and `0007_turn_triage.sql` are committed, idempotent, and were never registered**, so `db:migrate` skipped them. Any fresh deploy from migrations was missing three `users` columns and four `turn_ledger` columns plus their four indexes — the columns Coach Health's triage view reads and the durable client facts the coaching path reads. `db:push` had been hiding it, which is precisely what that guard warns about.

**No new SQL.** I generated a migration first, saw it duplicated the two orphans, deleted it, and registered the files that already existed. The migrations were written and committed; only the journal disagreed. Both are `IF NOT EXISTS`, so registering them is safe where a push already applied them.

Proved by building a database from empty with `db:migrate` and running the whole acceptance script against it.

### 2 · The adjacent-day corruption

The composition branch selected `.where(userId).orderBy(desc(loggedAt)).limit(1)` — the globally newest meal — and consulted `parseMealDate` only for `target`, and only when `plan.moves`. So `"Monday wasn't toast, it was rice."` edited whatever was logged last: Monday kept its toast, **Wednesday** gained the rice.

An explicitly named day now constrains the candidate there too. **A move is deliberately left alone** — that is the one case where the named day is the destination, not the source.

`namedPastDay` answers "which day did they name" once for both correction branches in this file, replacing the copy the Tuesday fix had introduced. No new owner.

`monMoved` and `wedIntact` are **blocking assertions** now, not a report:

```
Mon 220/5  [Rice]                corrected=true   <- their own row
Tue 644/46 [Chicken thigh, Pap]  corrected=true   <- rice replaced, chicken kept
Wed 593/51 [Pap, Ox liver]       corrected=false  <- byte-identical
```

Tuesday also improved as a side effect of correct targeting: the composition branch now replaces rice **within** the composite and keeps the chicken, instead of the identity path collapsing the row to pap alone.

### Verification

- **pg-correction-acceptance: GREEN, 14/14**, PostgreSQL **16.13**, database built from **committed migrations only**
- **`npm test`: all suites green — 36/37**, one honestly vacuous (normalizer replay, no corpus). `check-architecture` is green on this base — #165 closed 32/21/423.
- schema-safety **OK** · file-size 236 files OK · names 97/97 · reach 8 · SAST 61/61
- Gauntlet unchanged from main: Slices 1/2/4 green, 3 and 5 red on the same pre-existing sentence-length findings
- Red on revert still bites: today-only window fails four checks; an unconstrained newest-row select fails the two adjacent-day ones

**Handoff** — BASE `8c8cce1` / BRANCH `proof/pg-correction-164` / HEAD **`df2b0c5`** / PR #166 / MERGED no / CI re-triggered.

One thing for your judgement rather than mine: the journal gap means **any environment provisioned from migrations since those two files landed is missing seven columns**. Worth checking production before this is assumed to be a CI-only fix.

---

Issue #203, opened 2026-09-06
P1: sparse/intermittent clients must not collapse to receipt-only HOLD


ButiMM:
## CTO ACTIVE CLAUDE LANE — BOUNDED PRODUCT/POLICY CONVERGENCE

Baseline: current `origin/main` (at creation: `aa2fda8744daf6dc21858cc4cbac96399faecda2`).

This issue is the authoritative bounded job for the Claude builder lane. It is NOT a tracker and must not expand into a broader architecture programme.

## Founder product law

KamLife is a behavioural coach, not a tracking app.

Daily reporting is not a prerequisite for useful coaching. First-class customer behaviour includes:
- daily logging;
- reporting every 2–4 days while still executing;
- multi-day catch-up messages / voice notes;
- disappearing and returning;
- partial recollection;
- executing without reporting;
- financially constrained weeks.

Do not equate `no recent log = no adherence` and do not equate `missing evidence = say nothing`.

When evidence is insufficient to prescribe safely, Coach K should ask the smallest high-information question that materially changes the next decision — unless the existing reply/card/question already owns NEXT.

## Current-main evidence already established by CTO

`closeCoachingTurn()` now calls the canonical decision after real durable writes and `durableDomains()` recognizes food, step INSERT/UPDATE, workout, weight and water writes. Do NOT reopen old stale claims that water or step updates are invisible.

The live product-policy divergence is higher:

- `canonicalDecision()` delegates to `chooseAction → underPolicy`.
- `underPolicy` can downgrade a thin-evidence client to `hold`.
- `withNextMove()` explicitly treats an empty todo from an unevidenced client as "the correct silence" and appends nothing.
- Result: a truthful receipt can remain only a receipt because the policy decided that insufficient evidence means no coaching action/question.

This is in tension with the founder contract above and with proactive behaviour that can investigate under thin evidence.

## Job

Reproduce this on current `main` through the real front door and real PostgreSQL where state matters.

Use at minimum:

1. New/sparse client logs a normal meal: `I had pap and chicken`.
2. Sparse client reports steps: `I walked 8000 steps today`.
3. Sparse client logs a weigh-in.
4. Client returns after several quiet days and reports one fact.
5. Client reports several days at once.
6. Client has insufficient evidence for a safe prescription but enough context to ask one useful question.
7. Control: reply/card already owns NEXT — do not append another move/question.
8. Control: evidence is sufficient and a real next action exists — preserve the existing action.
[...]

PR #205, opened 2026-09-06
#203: a sparse client is coached, not just receipted — DO NOT MERGE before CTO adjudication

ButiMM:
**`FIXED + PROVEN`** for #203. Baseline `origin/main@aa2fda8744daf6dc21858cc4cbac96399faecda2` (the SHA the issue names). Branch HEAD `e8c5bc27c436e498950dc9d40904a191acec8166`.

**Held at CTO adjudication — not merged, not deployed**, per the issue's stop condition.

## The first divergence is one rung earlier than the issue predicted

Traced through the real front door on real PostgreSQL before anything changed:

```
"I walked 8000 steps today"   steps met, rest day, weighed today, 1/7 days logged
chooseAction  -> hold          every rung stands down, each of them correctly
underPolicy   -> hold          never reached: `hold` is not PRESCRIPTIVE
final reply   -> "8 000 steps — nice one. 👌"     and nothing else
```

The issue names `underPolicy` downgrading a thin-evidence **prescription** to `hold`. That happens. But for a sparse client the ladder holds far more often than it prescribes, so the commonest road to the dead end was the one the guard did not cover.

My own first trace missed it too — a hand-built `DayState` used `sessionsTarget: 3` where `live.ts` zeroes it on a rest day, which put the client on the `train` rung and hid the real path. Written down in the acceptance file so the next reader does not repeat it.

### 1. The gate held where the other path investigated

`decideProactive` already downgrades an unevidenced prescription to the measurement that would justify one. `underPolicy` went to `hold`.

`one-action.ts` carried this divergence written down as **FOUND, NOT FIXED — parked 2026-08-28** because the product question was unproven, and it named the exact close: *"this gate also performing the askToLog/askToWeigh downgrade, which needs today.logged, daysSinceWeighIn and doNotMention at the reactive call sites."*

Both halves are now settled. #203's founder rule answers the product question. And the three facts were never missing — every call site already computes them into the `DayState` it hands to `chooseAction`; they were simply not passed on.

The downgrade is **lifted, not copied**: one `investigateInstead`, called by `decideProactive` and by the gate. A `hold` under **sufficient** evidence is an earned verdict and is untouched — *"the most common correct decision may be CONTINUE"*. A `hold` under **insufficient** evidence is a different sentence wearing the same word: *"we cannot tell, and we did not ask."*

### 2. A same-day re-weigh was invisible

#147 made the second weigh-in of a SAST day **UPDATE** the row instead of inserting. `durableDomains` still matched only `/INSERT weight/`, so the turn read as no durable write at all, `closeCoachingTurn` returned before the decision owner, and the client got a bare receipt.

```
first weigh-in of the day   INSERT weight  ->  "87.4kg — noted. 👌
                                                Tell me what you ate today — one line is enough."
second, same day            UPDATE weight  ->  "87.4kg — noted. 👌"
```

Same client, same sentence, same truth written. One pattern in the same table, matching `steps` and `water`, which already carry their UPDATE verb. This is not the water/steps claim the issue told me not to reopen — those are fixed; weight was not.

## Customer-visible outcome

| scenario | before | after |
|---|---|---|
| steps from a sparse client | `8 000 steps — nice one. 👌` | + `Tell me what you ate today — one line is enough.` |
| weigh-in from a sparse client | `87.4kg — noted. 👌` | + the same question |
| second weigh-in that day | receipt only | coached identically to the first |
| three days at once | the 3-day receipt | + the question |
| evidenced client | `protein` action | **unchanged** |
| logged today already | — | **never asked to log again** |

## Controls — what must not change, each asserted

- an evidenced client keeps the real action the ladder chose (`kind === "protein"`);
- a client who logged today is never asked for the meal they just logged;
- a client who ruled out the scale is never asked to weigh;
- a closed food day is never answered with an instruction to eat;
- the question appears **at most once** — `withNextMove` still owns whether the reply already carries NEXT;
- reactive and proactive, on the same rows, both investigate rather than one holding.

**The degraded proactive call sites pass no context on purpose.** Their `loggedToday`/`daysSinceWeighIn` are placeholders for a ledger read that just failed, not facts; handing them to the downgrade would ask a client to log off a value we invented. Omitting the context reads as "nothing worth asking" and holds — exactly what those branches did before.

## Red on revert — both mechanisms, independently

```
gate stops investigating a ladder hold   FAIL  a step report … ends in the question
                                               "8 000 steps — nice one. 👌"
                                         FAIL  the reactive gate investigates rather than holding
                                               kind=hold

weight UPDATE pattern narrowed           FAIL  a weigh-in … ends in the question
                                               "87.4kg — noted. 👌"
                                         FAIL  the second weigh-in of a day is coached like the first
```

## One source-shape assertion regraded, not restored

`gap-tests` grepped `one-action.ts` for `mentionsForbidden("weight scale weigh", p.doNotMention)`. The guard moved into the shared ladder and now reads `ctx.doNotMention` — same guard, same rule. Restoring the old string would preserve a shape rather than the promise, so the **outcome** is asserted in its own test, with a positive control proving the same state *does* reach the scale when nothing is forbidden. That catches a guard deleted rather than renamed, which the grep never could.

## Verification

- `npm run check` — clean.
- `npm test` — **37/37 green**.
- Journey Lab — GREEN, 266 checks across 8 sections.
- All **ten** PostgreSQL acceptance suites GREEN, including the new `pg-thin-evidence-acceptance` (16 checks), wired into the `pg-acceptance` job.
- No governor raised, no assertion weakened, no new response-authoring system, no parallel question engine, no new policy ladder.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,6 +155,12 @@ jobs:
       # shadow door. Neither is expressible on a fixture that answers every query the same way.
       - name: One weight-direction authority on real PostgreSQL
         run: npx tsx script/pg-weight-authority-acceptance.ts
+      # A SPARSE CLIENT IS COACHED, NOT JUST RECEIPTED (#203). Every decision here is computed
+      # from rows — how many days carry a meal, whether one carries TODAY, how stale the weigh-in
+      # is, and whether a same-day re-weigh updates or inserts. None of that is expressible on a
+      # fixture that answers every query the same way.
+      - name: Thin-evidence coaching on real PostgreSQL
+        run: npx tsx script/pg-thin-evidence-acceptance.ts
       # THE SIX JOURNEYS (#170). Same database, same migrations, same front door — a second job
       # would be a second copy of this infrastructure for no gain. Deliberately NOT
       # continue-on-error: the whole point of the lab is that a known-wrong durable state cannot
--- a/script/gap-tests.ts
+++ b/script/gap-tests.ts
@@ -4361,12 +4361,42 @@ test("cut8: the decision stands down before the mouth ever has to", () => {
   assert.ok(/const scaleIsOffLimits = mentionsForbidden\("weight scale weigh", s\.doNotMention\)/.test(code),
     "the weigh ask is never chosen");
   assert.ok(/!scaleIsOffLimits && \(\(neverWeighed/.test(code), "…in the ordering itself");
-  assert.ok(/&& !mentionsForbidden\("weight scale weigh", p\.doNotMention\)/.test(code),
-    "…and in the verdict downgrade, which reaches askToWeigh by a second route");
+  // …AND IN THE VERDICT DOWNGRADE, WHICH REACHES askToWeigh BY A SECOND ROUTE.
+  //
+  // This used to grep for `mentionsForbidden("weight scale weigh", p.doNotMention)` in this file.
+  // #203 lifted that downgrade out of decideProactive into one ladder both paths call, so the
+  // guard now reads `ctx.doNotMention` — same guard, same rule, a different local name. Restoring
+  // the old string would preserve a stale shape instead of the promise, so the OUTCOME is asserted
+  // in its own test below: it survives the move, and it would also catch a guard deleted rather
+  // than renamed, which the grep never could.
+  assert.ok(/mentionsForbidden\("weight scale weigh"/.test(code), "the downgrade consults the boundary");
   // And the fact has to actually arrive: morning carries it into the profile it decides from.
   const morning = readFileSync("server/scheduler/jobs/morning.ts", "utf-8");
   assert.equal((morning.match(/doNotMention: client\.doNotMention/g) || []).length, 3,
     "all three decision call sites in morning carry it");
+});
+
+test("cut8: the verdict downgrade never asks for a scale the client ruled out", async () => {
+  const { chooseAction, underPolicy } = await import("../server/one-action");
+  // Sparse, and never weighed: exactly the conditions under which the downgrade reaches the scale.
+  const state: any = {
+    goal: "fat_loss", weeksOnProgramme: 5, daysSinceAnyLog: 1, daysSinceWeighIn: null,
+    loggedToday: true, proteinPct: 1, caloriePct: 1, sessionsThisWeek: 0, sessionsTarget: 3,
+    stepsToday: 8000, stepsTarget: 8000, hour: 14, atKeyboard: true,
+  };
+  const thin = {
+    foodSufficient: false, weightSufficient: false, dreamGoal: null,
+    loggedToday: true, daysSinceWeighIn: null as number | null,
+  };
+  const raw = chooseAction({ ...state, doNotMention: null });
+  // THE CONTROL FIRST. Without the boundary this state DOES reach the scale, so the refusal below
+  // is a refusal rather than an accident of the fixture.
+  assert.equal(underPolicy(raw, { ...thin, doNotMention: null }).kind, "weigh",
+    "the downgrade should reach the scale for a never-weighed sparse client");
+  const banned = underPolicy(chooseAction({ ...state, doNotMention: "weight" }),
+    { ...thin, doNotMention: "weight" });
+  assert.notEqual(banned.kind, "weigh",
+    `the downgrade asked a client who ruled out the scale to weigh: "${banned.todo}"`);
 });
 
 // ── CUT 9: ONE FOOD-CONSTRAINT OWNER, AND THE WEIGHT REPORTS HONOUR THE COLUMN ──────────────
--- a/script/pg-thin-evidence-acceptance.ts
+++ b/script/pg-thin-evidence-acceptance.ts
@@ -1,0 +1,212 @@
+/**
+ * REAL-POSTGRESQL ACCEPTANCE — a sparse client is coached, not just receipted (#203).
+ *
+ * THE FOUNDER RULE THIS ENFORCES. KamLife is a behavioural coach, not a tracking app: reporting
+ * every few days, catching up in one message, disappearing and returning are all first-class
+ * customer behaviour. Insufficient evidence must not become a receipt-only dead end — when we
+ * cannot safely prescribe we ask the smallest question that would change the next decision, unless
+ * the reply already owns NEXT.
+ *
+ * TWO DEFECTS, BOTH FOUND BY TRACING THE REAL FRONT DOOR BEFORE ANYTHING WAS CHANGED.
+ *
+ *   1. THE GATE. underPolicy downgraded an unevidenced PRESCRIPTION to `hold`, and left a `hold`
+ *      that the ladder itself had reached untouched. For a sparse client the ladder holds far more
+ *      often than it prescribes — steps met, rest day, weighed today, one day logged in seven, and
+ *      every rung correctly stands down — so the commonest path to the dead end was the one the
+ *      guard did not cover. decideProactive already downgraded to the measurement that would
+ *      justify a prescription; the reactive gate now applies that same ladder.
+ *
+ *   2. THE WRITE THAT WAS INVISIBLE. A second weigh-in on the same SAST day UPDATEs the row rather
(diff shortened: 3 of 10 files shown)


ButiMM:
## CTO ADJUDICATION — BLOCKING COMMENT

**Do not merge #205 yet.** The architecture direction is accepted, CI is green, and the shared `investigateInstead` shape is the right convergence. One customer-visible blocker remains.

### BLOCK — `ProgressWeight.known` is being used as weigh-in recency

The new reactive call sites pass:

`daysSinceWeighIn: truth.weight.known ? 0 : null`

and #203 now uses that value to decide whether to ask the client to weigh.

But the canonical owner defines `known` as `change !== null` — whether a multi-point weight change can be computed — **not whether/when the client last weighed**. With one weigh-in, even one from this morning, `known === false` while `currentKg` and `points[0].at` are real. With two old weigh-ins, `known === true` even if the latest is stale.

So the new policy can produce both wrong directions:

- **single weigh-in today + logged today + sparse history** → passed as `daysSinceWeighIn=null` → may ask them to weigh again;
- **two stale weigh-ins** → passed as `daysSinceWeighIn=0` → stale scale evidence masquerades as current.

This approximation existed before, but #203 makes it customer-facing investigation policy. It must now be real recency.

### Smallest close

Do not create another weight reader. Derive actual `daysSinceWeighIn` from the canonical weight owner's returned latest point/timestamp, or expose recency from that existing owner if cleaner. Preserve `doNotMention` and fail-closed behaviour.

Required real-PG controls:

1. single weigh-in **today** + logged today + sparse history → **must not ask to weigh again**;
2. single genuinely **stale** weigh-in + logged today + sparse history → scale ask allowed when not forbidden;
3. **two stale weigh-ins** → must not masquerade as `0 days`;
4. `doNotMention=weight` still prevents the ask;
5. food-first investigation and evidenced-action controls remain green;
6. red-on-revert specifically for the recency mechanism.

Everything else in this cut remains accepted in direction. Fix on the same branch/PR and return to the CTO gate.

ButiMM:
## CTO CONTINUATION — ADDRESS BOTH LIVE BLOCKERS ON THIS PR

Stay on PR #205 / branch `fix/thin-evidence-investigate-203`. Do not open a new PR and do not merge.

The architectural direction remains accepted, but two customer-visible defects must close before approval:

### 1. Real weigh-in recency, not `ProgressWeight.known`

The existing CTO blocking comment stands. `truth.weight.known ? 0 : null` is not recency. Use the existing canonical weight owner/timestamps to derive actual days since latest weigh-in. Preserve doNotMention and fail-closed behavior. Required real-PG controls already listed in the CTO block remain mandatory.

### 2. Proactive HOLD must use the same investigation downgrade

Codex review found a second convergence hole: `decideProactive` still invokes `investigateInstead` only inside `if (PRESCRIPTIVE.has(action.kind) && evidence === "insufficient")`. A sparse client whose `chooseAction` result is already `hold` therefore still returns proactive `CONTINUE`/empty while reactive `underPolicy` now converts the same insufficient-evidence hold into an investigative ask.

That violates #203's explicit reactive/proactive convergence contract.

Smallest close: make the proactive path consult the same existing investigation owner for an **insufficient-evidence hold** as the reactive path, without turning sufficient-evidence HOLD/CONTINUE into nagging. Do not add another ladder or question engine.

Required controls:
- same sparse rows produce the same investigative kind reactively and proactively;
- sufficient-evidence HOLD remains HOLD/CONTINUE on both paths;
- logged-today client is not asked to log;
- fresh weigh-in is not asked again;
- stale/never-weighed scale ask obeys doNotMention;
- closed food day / declined training constraints still stand;
- at most one NEXT/question reaches the reply;
- red-on-revert independently proves the proactive-hold mechanism.

After both fixes: run the existing #203 real-PG acceptance, Journey Lab, full required local suite, push the same branch, and return to CTO gate. Do not merge/deploy.

ButiMM:
Both blockers closed at `f5733cb`. Same branch, no merge.

## Blocker 1 — `known` was being read as recency

`getWeightTruth` — the existing canonical weight owner — now returns `daysSinceWeighIn` beside `known`. No second weight reader was created; the rows and their instants were already in hand there. The three call sites (`understanding/live.ts`, two in `handlers/misc-commands.ts`) read it instead of computing `known ? 0 : null`. `doNotMention` still stands the whole read down before the query and returns `daysSinceWeighIn: null`; the degraded fallback sites in `scheduler/jobs/morning.ts` and `scheduler/proactive-decision.ts` deliberately pass no investigation context, since the zeros there are placeholders for a failed ledger read, not facts.

**One extra defect surfaced while proving it.** A weight row can carry a usable kg and an unusable instant — `points` only ever filtered on `Number.isFinite(kg)`, so a missing `loggedAt` survived as an Invalid Date. Latent while nothing read it; a thrown `RangeError` inside the SAST helper the moment recency did, taking the whole `getProgressTruth` read down and dropping three answers to the GPT fallback. Recency is now graded on a usable instant and reads `null` — "I do not know when" — rather than throwing. Same fail-closed contract as never-weighed and withheld.

**Red-on-revert (recency owner removed → `change !== null ? 0 : null`):** 4 checks red, reproducing both wrong directions named in the blocker.

```
FAIL  a client who weighed TODAY, once, is not asked to weigh again
      "8 000 steps — nice one. 👌\n\nStand on a scale this morning, before you eat."
FAIL  two stale weigh-ins do not masquerade as weighed-today
FAIL  …and the owner reports the real figure, not the `known` flag
      known=true daysSinceWeighIn=0
```

**Red-on-revert (unusable-instant guard removed):** `tracking-contract-tests` 4/72 violations — three GOAL_DISTANCE answers collapse to `"Sorry Kam, I didn't quite catch that"` and a durable weight write ends in a bare receipt. Green with the guard.

## Blocker 2 — the proactive path holds where the reactive path investigates

`decideProactive` now consults the same `investigateInstead` owner for an insufficient-evidence `hold`, not only for a `PRESCRIPTIVE` action. No second ladder, no new question engine. A hold under **sufficient** evidence is still an earned verdict on both sides — controlled explicitly.

**Red-on-revert.** The first attempt at this control did not bite, and that is worth recording: proactively `atKeyboard` is false, so `come_back` fires on three days of silence and hides the hold behind a real rung — the suite stayed green with the mechanism removed. The fixture was rebuilt to actually reach the proactive ladder's own hold: logged **yesterday** (no come-back), steps walked, week's sessions done, weighed this morning. Every rung stands down under one day of evidence in seven.

With the mechanism reverted to `PRESCRIPTIVE.has(action.kind) && evidence === "insufficient"`:

```
FAIL  a proactive hold under insufficient evidence becomes the question instead of silence
      kind=hold todo="Nothing new today. Do exactly what you did yesterday."
FAIL  …and it is the food question, because that is the measurement actually missing
      kind=hold todo="Nothing new today. Do exactly what you did yesterday."
FAIL  …and it becomes a real line rather than an empty proactive send
      ""
```

The third failure is the customer-visible symptom: `line=""`. Not a bad message — no message.

One assertion of mine was corrected rather than the product: asserting an identical `kind` across both paths failed with `reactive=log proactive=come_back`. That is the documented `atKeyboard` difference — the proactive side skips `come_back` for a client typing right now. The check now grades what actually has to agree: both investigate, neither holds, both reach for the same missing measurement.

## Runs

| Suite | Result |
|---|---|
| `pg-thin-evidence-acceptance` (real PG) | **GREEN** — 27 checks |
| All 10 pg acceptance suites (real PG) | **GREEN** |
| `npm test` | **GREEN** — 37/37 |
| Journey Lab | **GREEN** — 266 checks across 8 sections |

No budget raised, no assertion weakened, no exclusion added, no `continue-on-error`.

Awaiting CI, then back to your gate. Not merging.

---

ButiMM:
**Remote CI green on `f5733cb`.**

Run [782](https://github.com/ButiMM/KamLife-Coach/actions/runs/34041126243) — `conclusion: success`, all three jobs:

- `unit-baseline` — success (unit failures compared to exact merge base)
- `test` — success (type check, full suite, gauntlet)
- `pg-acceptance` — success, every step including **Thin-evidence coaching on real PostgreSQL** and **Six critical journeys through the real system**

Both blockers closed. At your gate — not merging, not deploying, no new PR.

---

Issue #213, opened 2026-09-07
CODEX STRUCTURAL CUT 1C: information-value investigation through the existing canonical decision owner


ButiMM:
## CTO ACTIVE CODEX / GPT-5.6 SOL CUT — TRANCHE 1C

Baseline: newest merged `origin/main` after #212 — expected `9156d471e898cb2c323c4eb97a77219cf3856d32`. Fetch and record the actual SHA before work.

This is the next bounded post-sweep structural cut from #209. Do NOT restart the whole-system sweep. Do NOT revisit #208 except where its newly merged open-loop state is a dependency.

## Product law

**Truth is the foundation. Coaching is the product. Behaviour change is the outcome.**

KamLife must work for intermittent humans. Tracking compliance is not the product.

# Customer failure

When the canonical decision does not have enough evidence to justify a confident action, KamLife often collapses to a fixed `log food` / `weigh yourself` investigation. That is safe, but not yet intelligent coaching.

A long-term coach should ask the **single missing question whose answer is most likely to change the next decision**.

The product goal is NOT a generic self-questioning agent. It is one bounded extension of the existing INVESTIGATE authority.

# Canonical journeys to reproduce first

Use current main through the real customer/front-door paths and real PostgreSQL where state/evidence matters.

### A — weight stalled, food evidence thin
If the system cannot safely decide whether to change calories because weekend adherence is unknown, a high-value question may be about the weekend rather than another generic log request.

Expected product shape:
`insufficient decision evidence -> one highest-value missing fact -> one concise question -> answer changes/reconfirms the canonical decision`

### B — training progress unclear but an open coaching loop exists
The newly merged #208 unresolved ask must outrank creating a competing investigation question. Do not ask a new question while a material open coaching loop is already awaiting outcome unless the existing authority explicitly makes that safe.

### C — adequate food evidence, stale/missing weight
When weight is the actual missing measurement that blocks a decision, the existing weigh investigation should remain correct. This cut must not make simple cases cleverer for no reason.

### D — adequate weight evidence, unknown execution/context
When the numeric evidence is sufficient but a behavioural/context fact would materially change the action, investigate that missing fact rather than reflexively asking for another measurement.

### E — no evidence that a question would change the decision
Hold or give the existing safe action. Do not interrogate the client merely because the system can ask a question.

# First job — reconstruct the existing INVESTIGATE authority

Trace current main for:
[...]

Issue #211, opened 2026-09-07
CLAUDE CUT: facts, retractions, and same-turn propagation on current main


ButiMM:
## CTO ACTIVE CLAUDE BOUNDED CUT

Baseline: newest merged `origin/main` — currently `8f88e2bff9ece40e3cadd522f242f029378b7d87` after #207/#210.

This is **Tranche 2A** from #209, extracted from the completed whole-system sweep. Codex/Sol is concurrently active on #208 open coaching loops. **Do not touch #208's cross-turn open-loop ownership or any new coaching-loop state.**

## Product problem

KamLife must preserve what the client actually said and make a committed fact affect the same customer turn. The sweep/reviews previously identified cases where a fact, retraction, replacement, question, and correction in the same message could diverge between persistence, canonical belief, and the actual reply.

This cut is not a new parser project. Reproduce the current-main behaviour first and fix only the first divergence through existing owners.

## Mandatory current-main reproductions

Use the real production front door and real PostgreSQL where persistence/order matters.

### Fact + question in one natural sentence
- `I'm vegan now, what should I eat?`
- `I can't eat dairy, what can I eat?`
- `my knee is killing me, what should I train?`

For each prove:
`client words → fact extracted → durable truth → canonical belief → answer owner → actual reply`

The committed fact must affect the **same customer-visible reply**, not merely a database column that becomes useful next turn.

### Retraction / replacement
- `I'm not vegan, I'm vegetarian`
- `I'm not vegan anymore`
- injury language containing `anymore`, including a question where the user is not retracting the injury

Prove that replacement/retraction semantics are exact and chronology-safe.

### Exact removal
- retracting `nuts` must not erase an unrelated `peanuts` allergy/restriction;
- one specific fact removal must not broad-match a different fact;
- unrelated durable facts must remain untouched.

### Mixed-intent / separator control
Reproduce at least one mixed fact + question with no comma or clean separator. A question elsewhere in the message must not veto a stated fact elsewhere in the message.

## Ownership trace

Before changing code, locate the existing owners for:
- fact extraction/classification;
- durable user fact write/revision;
- retraction/replacement semantics;
- canonical user/context read;
- same-turn reply context/decision;
- food/training constraints where those facts affect the reply.

Return the first divergence for every failing journey. Do not start at the final bad sentence if truth was already lost earlier.

## Architecture constraints

Do NOT:
- create a second fact parser;
- create a second memory store;
- add a generic correction/retraction engine;
[...]

Issue #208, opened 2026-09-06
CODEX STRUCTURAL CUT: one open coaching loop across turns without a second policy system


ButiMM:
## CTO ACTIVE CODEX / GPT-5.6 SOL STRUCTURAL CUT

Baseline: newest merged `origin/main` (currently `84601ba90d90d180059527c20afc74428367f939`).

This starts **after** the original whole-system Sol sweep/checkpoint completed locally. Do not restart #204. Use what you learned there as context, but re-verify every implementation claim against current main.

Claude is actively executing #207 on durable-log response coherence. **Do not modify or compete with #207's response-authoring seam.** You may read it only to understand boundaries.

## Why this cut exists

Founder/manual-client reality and the independent Grok red-team agree on the missing behavioural spine:

- clients do not report perfectly every day;
- some execute and report 2–4 days later;
- some disappear and return;
- some tell Coach K they did what it asked without having logged it at the time;
- Coach K currently has a strong canonical truth/decision spine but weak durable state for an unresolved coaching ask and its later outcome.

KamLife must optimise for **behavioural continuity**, not tracking compliance.

The target is not a generic agent framework. The target is ONE real coaching loop that survives across turns:

`Coach K makes a concrete next move → client later reports outcome → system attributes that outcome to the open move → resolves/updates the loop → next decision sees the result`

## Canonical customer journey to prove

Use a training commitment as the first end-to-end proof because it is explicit and objectively resolvable:

1. Coach K tells a client to complete today's/next training session through the existing canonical decision owner.
2. The client does not send a structured completion at that moment.
3. Later — same day or next day — client says naturally:
   - `I did the workout you told me to do`
   - `I trained yesterday, forgot to tell you`
   - `I got the session done`
4. KamLife must:
   - preserve the client's actual words and chronology;
   - attribute the report to the unresolved coaching move when evidence supports it;
   - use the existing workout/session truth owner for any durable completion that should exist;
   - resolve the open coaching loop exactly once;
   - prevent the next decision/proactive path from nagging the client to do the action they already completed;
   - preserve uncertainty when the wording does not identify which action was completed.

Negative journeys:
- `I couldn't do it, work was chaos` must NOT create a workout completion;
- `I might train later` must NOT resolve the loop;
- `what workout did you tell me to do?` is a question, not completion;
- an old unrelated coaching move must not capture a new completion;
- two users and two days must remain isolated;
- a real structured workout completion must not create a duplicate completion when it also closes the loop.

[...]

Issue #206, opened 2026-09-06
CLAUDE NEXT CUT: current-main food identity + portion reconciliation


ButiMM:
## CTO ACTIVE CODEX / GPT-5.6 SOL BOUNDED CUT

This replaces the broad #204 execution run. #204 is reference-only.

Start from the newest merged `origin/main`. Record the exact SHA. Read applicable `AGENTS.md` completely.

Do NOT touch Claude's active #203/#205 policy-convergence files while that lane is open.

## Customer problem

Recent PR #200 merged food-identity fixes, but merge prose is not proof that current customer behavior is fully closed. Reconcile the actual product through the real front door and real PostgreSQL where persistence matters.

## Required reproductions

1. `pap and a beef stew` must NOT invent `Pap en vleis`.
2. `toast and a butter chicken curry` must NOT invent `Toast with butter`.
3. `half a gatsby` must preserve the HALF Gatsby identity and portion.
4. `half gatsby` must remain equivalent to the intended half-portion alias.
5. rice + chicken livers must remain separate foods in both word orders.
6. Kota identity must remain deterministic and preserve the client's stated portion.
7. Gatsby identity must remain deterministic and preserve the client's stated portion.
8. Negative control: legitimate aliases and joins must still work; do not fix collisions by disabling useful recognition.

For every failure trace:

`client words → scanner/matcher identity → stored item(s)/portion → canonical day truth → actual customer reply`

Find the FIRST divergence.

## Execution authority

If a defect reproduces and the existing matcher/scanner/identity owner is clear, fix it in that owner, add the smallest customer-visible regression + negative controls, prove red-on-revert, and leave one clean local commit/PR for CTO adjudication.

Do NOT create a second parser, matcher family, food store, alias engine, or response layer.

If all required customer journeys pass on current main, return `NOT REPRODUCED` with exact evidence. Do not manufacture work.

## Evidence

Use the real product front door early. Use real PostgreSQL when the defect can affect stored identity/portion. Label evidence as LOCAL REAL POSTGRESQL / FIXTURE / MOCKED. A unit matcher test alone is not enough if the customer-visible route can still diverge.

## Stop condition

Return exactly one of:
- `FIXED + PROVEN` with branch, commit, PR, real journey evidence, negative controls, regression suites, red-on-revert;
- `NOT REPRODUCED` with exact current-main journey evidence;
- `DESIGN DECISION REQUIRED` only if a genuinely consequential owner/architecture decision is unavoidable.

[...]

PR #214, opened 2026-09-07
#211: facts, retractions and same-turn propagation — DO NOT MERGE before CTO adjudication

ButiMM:
**`FIXED + PROVEN`** for #211 / 2A. Baseline `origin/main@9156d471` (the SHA the start note names). Branch HEAD `3c44f02`.

**Held at CTO adjudication — not merged, not deployed.** No `#208` open-loop state touched, no `#213` question-selection authority touched, no food identity work.

Reproduced through the real front door on real PostgreSQL before changing anything. **Five defects**, each fixed at its own first divergence, all through owners that already existed.

## Acceptance A–H — before → after

| | client words | before | after |
|---|---|---|---|
| **A** | `I'm vegan now, what should I eat?` | column stored `"vegan now"`; constraints read out as **two** terms `[vegan, vegan now]` | column `vegan`; one term; reply vegan-safe |
| **B** | `I can't eat dairy, what can I eat?` | *"You said you are done eating today, so I am leaving it there. You finished on 0 kcal."* | a real, dairy-safe meal suggestion |
| **C** | `my knee is killing me, what should I train?` | already correct | **GREEN on current main — no code written** |
| **D** | `I'm not vegan, I'm vegetarian` | column `vegan, vegetarian`; reply *"updated to **vegan**"* | column `vegetarian`; reply *"updated to **vegetarian**"* |
| **E** | `I'm not vegan anymore` | reply *"updated to **vegan**. I'll only suggest plant-based proteins"* | *"you're not vegan anymore. I've taken that off your record."* |
| **F** | `I can eat nuts again` on `peanuts, nuts` | **both cleared** | `peanuts` intact |
| **G** | `my knee doesn't hurt as much anymore, should I still avoid squats?` | knee injury **erased** | injury intact, question answered |
| **H** | `I am vegan now what should I eat` (no separator) | already correct | GREEN; isolation and unrelated-fact preservation asserted |

## The five first divergences

**1 · The span, not the token.** The lazy capture runs to the next comma, so the restriction was stored as `"vegan now"` — which read out of `foodConstraints` as two terms and left every retraction matching a string the client never said. Narrowed to the restriction word, off the one list the module already keeps.

**2 · A restriction is not a closed day.** `foodDayIsClosed` claimed `"I can't eat dairy"`. The composed clause directly below it already carried the rule — *the cessation must apply to eating ITSELF; an adverb or a food after the verb means they described how they eat* — and this earlier, simpler clause never got it. Now scoped the same way, checked both directions.

**3 · Exact removal.** `removeFact` matched on **containment**, for a stated reason (spans like `"vegan now"` had to be reachable). The cost was a **peanut allergy deleted by a sentence about nuts** — the most dangerous write in the module. A word boundary keeps the reason and drops the cost, and legacy span rows are still cleared (controlled).

**4 · A retraction beats the same *item*, not the whole *field*.** The merge deleted any assertion of a field carrying a retraction anywhere in the bubble, so `"I'm not vegan, I'm vegetarian"` ended with **no dietary truth at all**. Underneath it, a second bug: retractions run before appends by design, but the append re-read the *original* row, so the retracted item came straight back and overwrote the retraction's own patch — the client corrected us and ended up **more** restricted.

**5 · `anymore` is a time marker, not a resolution claim.** A bare `\banymore\b` alternative meant any sentence naming a body part and containing that word amputated the injury. It now has to attach to a phrase claiming the injury is over; a hedged improvement is not resolution; and the retract branch carries the same `reporting` gate the assert branch always had — a question could not *add* an injury but could *delete* one, and that asymmetry ran in the unsafe direction.

## And the mouth speaks the one truth

`DIET_PREFERENCE_UPDATE` held a second opinion about diet — it re-read the raw message, tested `vegan` before `vegetarian`, and its trigger matches `i'm … vegan` with anything in between. `recordClientFacts` has already run at the door before any routing, so this mouth now renders the canonical column, announces only when the turn actually changed the record, confirms a retraction, and clears the stale `diet:` flag in `profileNotes` so the meal-plan reader cannot keep building vegan plans off a restriction that is gone.

## Red on revert — seven mechanisms, independently

```
removeFact boundary   -> the peanut allergy is deleted again
reporting gate        -> asking IF a knee resolves resolves it
resolution regex      -> a hedged improvement amputates the injury
merge rule            -> vegetarian is deleted by the vegan retraction
append ordering       -> "vegan, vegetarian"
token narrowing       -> "vegan now" stored, read out as two terms
closure scoping       -> a dairy restriction closes the food day
announcer             -> the coach confirms the diet they just dropped
```

The `reporting` gate did **not** bite on the first attempt — the resolution-regex fix alone covered the hedged sentence. Rather than keep unproven code I added the control that needs it (`"if my knee doesn't hurt anymore should I go back to squats?"` — a full resolution phrase inside a conditional the client has not claimed), and it bites now.

## Controls — all asserted

A real closure still closes the day · a genuine resolution still clears the joint and takes **only** that joint · a legacy span row is still cleared by its retraction · another client's restriction is untouched · unrelated durable facts survive a retraction of something they never had · no diagnosis is invented.

## One assertion regraded, not weakened

`gap-tests` asserted the column held `"not eating dairy"` — the raw span. It holds `"dairy"` now. The promise is that *a restriction stated in the negative is still a restriction*, so that is what is graded: recorded, non-empty, and reaching `foodConstraints`.

## A finding, not fixed here

`"I'm vegan now, what should I eat?"` is claimed by the diet announcer rather than the meal-plan owner, so it answers with the vegan protein list and *"Type meal plan"* instead of a Next Meal Suggestion. The reply is vegan-safe and the fact lands, so acceptance A holds — but this is a genuine claim-precedence tension between a fact-statement and a plate-ask. It behaves identically on current main. Left alone deliberately: routing/question-selection is `#213`'s lane.

*(It surfaced as a false Journey Lab failure first — my acceptance script was not cleaning `turn_ledger`, and Coach Health reads that table for the whole database. Fixed in the script; the product was never involved.)*

## Verification

- `npm run check` — clean.
- `npm test` — **37/37 green**.
- Journey Lab — GREEN, **266 checks** across 8 sections.
- All **twelve** PostgreSQL acceptance suites GREEN, including the new `pg-facts-acceptance` (30 checks), wired into the `pg-acceptance` job.
- **No governor raised** — `regexLiterals` *fell* to 447 and is locked in. No assertion weakened, no exclusion, no `continue-on-error`.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -171,6 +171,12 @@ jobs:
       # not move. Real rows, real front door, several turns.
       - name: Durable log turns are one coaching turn on real PostgreSQL
         run: npx tsx script/pg-log-turn-acceptance.ts
+      # FACTS, RETRACTIONS AND SAME-TURN PROPAGATION (#211). Retraction-before-append ordering,
+      # one operation building on another's patch, and unrelated columns left alone are all
+      # properties of a real transaction — a fixture that returns whatever it was handed cannot
+      # show any of them, which is how a retracted diet came back inside its own write.
+      - name: Facts, retractions and same-turn propagation on real PostgreSQL
+        run: npx tsx script/pg-facts-acceptance.ts
       # THE SIX JOURNEYS (#170). Same database, same migrations, same front door — a second job
       # would be a second copy of this infrastructure for no gain. Deliberately NOT
       # continue-on-error: the whole point of the lab is that a known-wrong durable state cannot
--- a/script/check-architecture.ts
+++ b/script/check-architecture.ts
@@ -62,7 +62,7 @@ const BUDGET = {
    * to the true figure in one deliberate commit. Until then this red line is the marker, and it is
    * the ONLY thing in this guard that is red — one red line means something, four never did.
    */
-  regexLiterals: 448,
+  regexLiterals: 447,
   /**
    * GUARD #13 — see unreachableExports above. Sixty-two capabilities cannot be reached by a
    * client message today. This budget is deliberately set THREE BELOW that, so this guard is RED
@@ -273,7 +273,7 @@ const RAISES: Array<{ key: keyof typeof BUDGET; from: number; to: number; date: 
       + "so 21 is the smallest truthful current baseline. FROM HERE IT FALLS ONLY.",
   },
   {
-    key: "regexLiterals", from: 318, to: 448, date: "2026-08-24 (fell to 448 on 2026-09-05)",
+    key: "regexLiterals", from: 318, to: 447, date: "2026-08-24 (fell to 448 on 2026-09-05, to 447 on 2026-09-07)",
     why: "NOT A RAISE — A CORRECTED MEASUREMENT, and the follow-up this budget's own comment "
       + "declared owed on 2026-08-17: \"repair the matcher to see multi-line assignments and "
       + "re-baseline to the true figure in one deliberate commit.\" This is that commit. The "
--- a/script/gap-tests.ts
+++ b/script/gap-tests.ts
@@ -4239,8 +4239,19 @@ test("#128/1: a client can take a restriction back", async () => {
     const kept = projectClientFacts({ dietaryRestrictions: "allergic to fish" }, detectFacts(said));
     assert.equal(kept.patch.dietaryRestrictions, undefined, `"${said}" leaves the allergy alone`);
   }
-  assert.equal(projectClientFacts({}, detectFacts("I'm not eating dairy")).patch.dietaryRestrictions,
-    "not eating dairy", "a restriction stated in the negative is still a restriction");
+  // A RESTRICTION STATED IN THE NEGATIVE IS STILL A RESTRICTION — and #211 narrowed WHAT gets
+  // stored, not whether. This asserted the raw span the capture happened to reach ("not eating
+  // dairy"); the column now holds the restriction word itself, because the span form put
+  // "vegan now" in the record and read back out of foodConstraints as two terms. Graded on the
+  // promise instead of the old string: the restriction is recorded, and it reaches the one reader
+  // that decides what may be offered.
+  const negative = projectClientFacts({}, detectFacts("I'm not eating dairy")).patch.dietaryRestrictions;
+  assert.equal(negative, "dairy", "a restriction stated in the negative is still a restriction");
+  assert.ok(foodConstraints({ dietaryRestrictions: negative } as any).terms.some(t => /dairy/i.test(t)),
+    "…and it reaches the constraint reader that governs every suggestion");
+  // …and the narrowing must not swallow the restriction entirely — a column that ends up empty
+  // would pass a "no longer says not eating dairy" check while starving the client of the guard.
+  assert.ok(negative && negative.length > 0, "the narrowed restriction is not an empty column");
 });
 
 // ── #128/4: THE PROTEIN CLOSER ──────────────────────────────────────────────────────────────
--- a/script/pg-facts-acceptance.ts
+++ b/script/pg-facts-acceptance.ts
@@ -1,0 +1,259 @@
+/**
+ * REAL-POSTGRESQL ACCEPTANCE — facts, retractions, replacements, same-turn propagation (#211).
+ *
+ * THE RULE THIS ENFORCES. What the client says about themselves must survive intact, and a fact
+ * committed by a message must shape the SAME customer-visible reply — not a column that becomes
+ * useful next turn. Five defects, each traced through the real front door before anything changed:
+ *
+ *   "I'm vegan now, what should I eat?"   stored the restriction as "vegan now" — the span the
+ *                                         capture reached, which read back out of foodConstraints
+ *                                         as two terms and left retractions matching a string the
+ *                                         client never said.
+ *   "I can't eat dairy, what can I eat?"  read as a CLOSED FOOD DAY: "You said you are done
+ *                                         eating today, so I am leaving it there."
+ *   "I'm not vegan, I'm vegetarian"       vegan removed AND put back by the append, vegetarian
+ *                                         deleted by the merge, and the reply said "updated to
+ *                                         *vegan*" — the client corrected us and we confirmed the
+ *                                         thing they had just denied.
+ *   "I can eat nuts again"                cleared an unrelated PEANUT allergy, by containment.
+ *   "my knee doesn't hurt as much anymore, should I still avoid squats?"
+ *                                         erased the active knee injury, off a bare `anymore`, in
+ *                                         a question about training around it.
+ *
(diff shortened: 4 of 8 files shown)


ButiMM:
**Remote CI green on `3c44f02`.**

Run [790](https://github.com/ButiMM/KamLife-Coach/actions/runs/34118074243) — `conclusion: success`, all three jobs (`unit-baseline`, `test`, `pg-acceptance`), including the new **Facts, retractions and same-turn propagation on real PostgreSQL** step.

At your gate — not merging, not deploying. Standing by for #206 after adjudication.

---

ButiMM:
## CTO GATE — TWO BOUNDED BLOCKERS ON #211

CI is green and the main direction is correct. **Do not merge #214 yet. Stay on this PR/branch. Do not start #206.** Two exact-semantics holes remain inside #211's contract.

### 1. BLOCK — the new same-bubble replacement merge still uses substring containment

`detectFacts` now correctly says a retraction should beat only an assertion of the **same item**, but the implementation decides sameness with:

`asserted.includes(taken) || taken.includes(asserted)`

That recreates the exact `nuts`/`peanuts` class that this cut just fixed in `removeFact`, one layer earlier.

Reproduction to add through the real front door:

`I'm not allergic to nuts, I'm allergic to peanuts`

Expected: `nuts` retracted, `peanuts` becomes current truth.

Current code path: merged assertion=`peanuts`, retract=`nuts`; `"peanuts".includes("nuts")` is true, so the new peanut assertion is deleted before projection. The boundary fix in `removeFact` never gets a chance to save it.

**Smallest close:** use the same exact/token-aware semantics for deciding whether an assertion and retraction name the same item. Do not add another parser or vocabulary. Add a decisive control proving a same-message `nuts → peanuts` replacement preserves peanuts, and red-on-revert must bite this merge rule.

### 2. BLOCK — a partial diet retraction can leave the old `profileNotes` diet flag alive

The announcer now reads canonical `user.dietaryRestrictions`, which is good, but stale `profileNotes` is only cleared in the branch:

`if (_takenBack && !_dietTruth)`

So if a client still has another valid diet truth after the retraction, the old derived flag survives.

Concrete control:

- starting durable truth: `dietaryRestrictions = "vegan, vegetarian"`
- starting derived note: `profileNotes = "diet:vegan"`
- client: `I'm not vegan anymore`
- expected: canonical truth is vegetarian **and every derived diet reader is vegetarian/not-vegan**

Today `_dietTruth` is `vegetarian`, `_assertedThisTurn` is empty, so `_newDietFlag` stays null and the retraction cleanup does not run. `utils.ts::_getPool()` still explicitly reads `profileNotes` and returns the vegan protein pool whenever `diet:vegan` remains. That means a successful canonical retraction can still leave a live customer-facing food mouth on the old diet.

**Smallest close:** on any diet retraction/replacement, reconcile the existing `diet:` profile-note flag to the canonical remaining dietary truth (or clear it when none remains). Announcement policy can stay separate; this is derived-state convergence, not a second diet owner. Add a real-PG control that proves the canonical column and a live derived protein suggestion agree after partial retraction. Red-on-revert must expose the stale vegan pool.

### Close conditions

- both reproductions red before the fixes and green after;
- existing A–H #211 acceptance remains green;
- `npm run check`, full test suite, Journey Lab, and all PG acceptance green;
- no new parser/store/owner, no #208/#213 changes, no budget raise;
- then return `FIXED + PROVEN` on this same PR and stop.

This is a micro-close, not a new lane.

ButiMM:
## `FIXED + PROVEN` — both gate blockers closed at `76cac15`

Same PR, same branch. **Not merged, not deployed. #206 not started.**

Remote CI green: run [791](https://github.com/ButiMM/KamLife-Coach/actions/runs/34121691442) — `conclusion: success`, all three jobs.

### Blocker 1 — same-bubble merge decided sameness by substring

Reproduced first, and it is **worse than the gate stated**: the column did not merely lose the replacement, it **emptied**.

```
"I'm not allergic to nuts, I'm allergic to peanuts"     column "nuts"
  before   detectFacts -> {retract:{dietaryRestrictions:"nuts"}}
           projected   -> {dietaryRestrictions: null}
           a client who had just declared a PEANUT allergy left with no restriction at all
  after    detectFacts -> {dietaryRestrictions:"peanuts", retract:{dietaryRestrictions:"nuts"}}
           projected   -> {dietaryRestrictions: "peanuts"}
```

**Smallest close, as directed.** The predicate is named once — `namesFact` — and both readers use it: `removeFact`, which takes an item out of a column, and the merge, which decides whether an assertion and a retraction name the same item. They can no longer disagree about what "the same item" means, which is precisely what let them drift apart. No new vocabulary, no second parser; the boundary rule that already existed, given one home.

### Blocker 2 — a partial retraction left a live mouth on the old diet

```
column "vegan, vegetarian" + profileNotes "diet:vegan" + "I'm not vegan anymore"
  before   canonical    -> "vegetarian"
           profileNotes -> still "diet:vegan"
           offered      -> "cooked lentils, soya mince, or firm tofu"   ← the VEGAN pool
  after    canonical    -> "vegetarian"
           profileNotes -> "diet:vegetarian"
           offered      -> "eggs, firm tofu, or cottage cheese"
```

Reconciled **where the canonical value changes** — inside `projectClientFacts`, off the value that same patch just produced — rather than in the one handler that happened to write it, so every path is covered. It is a projection of the column, not a second opinion: it decides nothing the column does not already say. Announcement policy is untouched.

### Controls added — neither fix may read as "stop removing" or "always clear"

- a same-message `nuts` → `peanuts` replacement keeps the peanut allergy, and `nuts` is the one that goes;
- a client who **is** still vegan keeps the vegan flag and the vegan pool through an unrelated peanut retraction;
- an unrelated profile note (`fish allergy`) survives the diet-flag reconciliation;
- a full retraction clears the flag and returns the omnivore pool;
- the live mouth is asserted, not just the flag — `proteinOptions` is read back off the row.

### Red on revert

```
merge sameness by substring   -> FAIL  a same-message `nuts` → `peanuts` replacement keeps
                                       the PEANUT allergy
flag reconciliation removed   -> FAIL  …and the derived diet flag no longer says vegan
                                 FAIL  …it says what the column says
                                 FAIL  …and the protein pool they are offered is no longer
                                       the vegan one
```

Both restored to GREEN immediately after.

### Close conditions

| | |
|---|---|
| both reproductions red before, green after | ✅ |
| existing A–H acceptance still green | ✅ — `pg-facts-acceptance` now **39 checks** |
| `npm run check` | ✅ clean |
| `npm test` | ✅ **37/37** |
| Journey Lab | ✅ **266/266** |
| all PG acceptances | ✅ **12/12** |
| no new parser/store/owner, no #208/#213 changes, no budget raise | ✅ — `regexLiterals` still 447 |

At your gate.

---

Issue #227, opened 2026-09-08
CLAUDE INFRA CUT: isolate every PostgreSQL acceptance without weakening the gate


ButiMM:
## CTO ACTIVE CLAUDE INFRASTRUCTURE CUT — START NOW

Baseline: newest merged `origin/main`, currently expected `bf80605fd51584464228c0c972900e1b65f56694`. Record the actual SHA before work.

This is Tranche 3B infrastructure/evidence from #209. It is deliberately independent of Codex #225/#217 product logic and of #221 journeys 4–7.

## Why this exists

The authoritative `pg-acceptance` workflow currently runs many PostgreSQL acceptances as sequential GitHub Actions steps against one shared ephemeral database.

That has caused TWO proven engineering-control failures:

1. **cross-acceptance contamination** — on #224, three supposed product regressions disappeared when each acceptance was rerun after its own database truncate;
2. **early-exit blindness** — while `pg-open-coaching-loop` is red, every later acceptance is skipped, including restriction consistency, behavioural-pattern proof, comeback-clock proof and Journey Lab. A branch can therefore have excellent local evidence while authoritative CI never executes the proof it added.

This has already wasted builder/CTO time and produced false diagnoses. Fix the test infrastructure, not the product.

## Required result

Keep ONE PostgreSQL service / ONE authoritative PG job, but make every acceptance isolated and observable.

The desired behaviour is:

`for each acceptance -> reset only the ephemeral test DB -> run acceptance -> record verdict -> continue to next acceptance -> after all have run, exit non-zero if ANY real acceptance failed`

So one red remains a red gate, but it no longer hides all later evidence and one suite cannot poison the next.

## Architecture / safety constraints

- NO `continue-on-error` that turns a real red into yellow/green.
- NO second PostgreSQL service/job merely to bypass ordering.
- NO reordering to make the current branch look green.
- NO production/runtime code changes.
- NO acceptance expectation weakening.
- NO database reset capable of touching a non-test database. The runner/reset path must fail closed unless it is demonstrably in the CI/test environment it is intended for.
- Preserve the current acceptance order unless there is a technical reason to centralise the list in one existing/new test-runner owner.
- Do not duplicate the acceptance inventory in multiple owners if it can be consolidated.

A single runner that owns: reset -> execute -> collect -> final exit is preferred if it is simpler and less error-prone than dozens of YAML `if: always()` clauses.

## Mandatory controls

[...]

PR #226, opened 2026-09-08
#221 journeys 1–3: one comeback clock, counted in SAST days

ButiMM:
Journeys 1–3 of #221. **Do not merge — at the CTO gate.** BASE `main@e81138e`.

Journeys 4–7 are traced and [externalised on the issue](https://github.com/ButiMM/KamLife-Coach/issues/221#issuecomment-5580746992) but **not touched** — the open-coaching-loop / final-training-move seam is held while Codex repairs it. Journey 5's evidence is banked for when it clears.

## Journey 1 — GREEN, recorded not rewritten

One welcome, and the next turn answers what was asked instead of greeting again. No change made.

## Journey 2 — the clock divided milliseconds

`daysSinceContact` was `Math.floor((now - lastActiveAt) / 86_400_000)` — elapsed 24-hour periods, not days. Pinned to 06:00 SAST on a Tuesday:

| last contact | elapsed count | SAST calendar | returning? |
|---|---|---|---|
| 23:30 Sunday | 1 | **2** | `false` |
| 23:30 Saturday | 2 | **3** | `true` |
| 12:00 Sunday | 1 | **2** | `false` |

**The decision flipped, not just a printed number.** `RETURNING_DAYS` is 2, so a client last seen 23:30 Sunday who writes 06:00 Tuesday was two days gone, counted as one, and was **denied the comeback entirely**. Evening-then-morning is what most gaps in this product actually look like — people go quiet at night and come back in the morning.

`sastDaysBetween` is the existing owner; day-ledger, one-action-command and the outcomes audit already count days through it. This module held the last second definition of what a day is.

**Found while proving it:** `daysSinceContact` did `new Date(String(x))`, so a caller holding epoch **milliseconds** got an Invalid Date and a silent `null` — read downstream as *"we have never heard from this client"* rather than as the number they passed. Date, ISO string and number all work now.

## Journey 3 — what they did while they were quiet was invisible

```
lastActiveAt      9 days ago
workout_logs row  3 days ago      <- mid-absence, real, durable

"about a week away — everyone has those stretches."
"🏋️ Training: *1* session in the 14 days BEFORE you went quiet"
```

Both wrong from one cause: the gap was measured from **contact** alone, so the one thing they did during the silence neither moved the number nor was placed correctly in the sentence. They read a coach that did not notice they trained. Now:

```
"3 days away — everyone has those stretches."
"🏋️ Training: *1* session in the last 14 days — including while you were quiet 👊"
```

[...]

PR #224, opened 2026-09-07
#220 recovery: close the rotation's own four regressions

ButiMM:
Fix-forward from `main@d8823c5`. **Do not merge — at the CTO gate.** No rollback: the earlier #220 work is correct and rolling back would reintroduce it.

Codex reviewed `b04283a` thirteen seconds before it merged. All four findings are real, and **three were introduced by the slot rotation and kosher change in that commit**. Every one reproduced against the real builder before anything changed.

## P1 · The week ate more than the list bought

Rotation concentrates the plan onto the surviving candidates; the shopping list stayed a fixed string sized for the plan the template *used to* produce. Measured on `vegan, lentils` at R300–600:

```
eats  1400g firm tofu        list bought  800g   (Firm tofu 400g×2)
eats   600g soya mince dry   list bought  250g   (Soya mince 250g dry)
```

The client shops on Sunday, follows the plan exactly, and runs out on Thursday — a worse failure than the one rotation fixed, because they did nothing wrong.

Quantities are now **derived**: the week's requirement is summed off the plan that was actually built, and each line re-rendered at however many packs that needs, price scaled to match. Nothing is invented — pack size and pack price both come from the line that was already there — and an item the plan never names is dropped rather than bought.

**Dry versus cooked is stated, not rounded away.** "150g cooked lentils" against a 500g bag of *dry* lentils is a 2.5× difference; treating them as one unit would under-buy by more than the defect being fixed.

The line is **edited in place** rather than rebuilt from parsed parts. Rebuilding dropped `dry` from `Soya mince 250g dry` — a client buying 750g by cooked weight buys nearly three times what they need — and it would have made this a second place that authors shopping-list text. It stays a quantity edit.

## P2 · A day-indexed schedule is not a candidate pool

For recomposition, `dinnerCarbs` is *"training days get a carb, rest days get none"* — the file's own comment says so. Every other array is a variety pool where any entry may land on any day; **I treated all of them alike.** Compacting and rotating this one printed, for a client excluding sweet potato:

```
Monday — Training Day 🏋️   Dinner: 100g beef mince + extra veg only (rest day) … 616 cal
```

The words "(rest day)" reaching them on a Monday, training-day calories still claimed over veg.

[...]

Issue #221, opened 2026-09-07
CLAUDE NEXT CUT 2E: one comeback clock and one training/session owner


ButiMM:
## CTO QUEUED CLAUDE CUT — START ONLY AFTER #220 IS ADJUDICATED + MERGED

Baseline when work starts: newest merged `origin/main`. Record exact SHA and read applicable `AGENTS.md`.

This is Tranche 2E from #209. Do not start while #220 is active.

## Product problem
KamLife must have one coherent continuity/session authority when a client disappears, returns, asks where they are in the programme, or reports training. Current/legacy surfaces have historically produced duplicate comeback welcomes, inconsistent inactivity clocks, contradictory session numbers, or more than one training recap/progression owner.

## Mandatory current-main reproductions
Use the real front door and real PostgreSQL where chronology/session ordering matters:
1. one returning client after a multi-day absence -> exactly one comeback/welcome behavior, not competing greetings;
2. the comeback clock must use one canonical last-meaningful-engagement owner and SAST chronology;
3. a client who executed/trained during the absence but reports later must not be treated as fully disengaged if canonical evidence says otherwise;
4. one training session recap/progression owner: after a logged session, all customer-visible surfaces agree on the current/next session;
5. reproduce the Journey Lab controls that currently tolerate more than one session number; the final customer turn must not expose contradictory session counts/numbers;
6. correction/backdated training evidence must update the same canonical session truth rather than creating a second progression history;
7. negative control: a single quiet day is not a comeback event; ordinary unrelated messages must not reset or fabricate progression.

Trace each failure:
`client chronology -> stored engagement/training events -> canonical comeback/session state -> decision -> exact customer-visible reply`

Find the first divergence. Record already-green journeys without rewriting them.

## Architecture constraints
- one comeback/engagement clock, not another re-entry store;
- one training/session progression owner, not a second counter;
- no new response author or duplicate welcome ladder;
- preserve #208 open coaching-loop continuity and the existing workout-state/training authorities;
- do not touch Codex pattern/catch-up architecture beyond consuming already-merged canonical state.

## Proof
Real PostgreSQL for chronology/session ordering; exact customer-visible positive + negative controls; red-on-revert for each changed mechanism; tighten any false-green Journey Lab assertion rather than weakening it; full CI/governors.

## Stop condition
[...]

PR #228, opened 2026-09-08
#227: every acceptance runs, every acceptance starts clean, and red is still red

ButiMM:
Tranche 3B infrastructure. **Do not merge — at the CTO gate.** BASE `main@bf80605f`.

No production or runtime code is touched: the diff is the workflow, two new files under `script/`, and one `.gitignore` line.

## What was wrong

The authoritative `pg-acceptance` job was **eighteen sequential steps sharing one ephemeral database**, and both failure modes in the issue are real and had already cost diagnosis time.

**Early-exit blindness.** A failing step ends the job. While `pg-open-coaching-loop` was red — step 10 of 19 — the nine behind it never executed. Measured on PR #226's own CI run:

```
correction … weight-authority        9 × GREEN
open-coaching-loop                   RED — 2      <- job exits here
thin-evidence, information-value, log-turn, facts, food-identity,
weight-speakability, restriction-consistency, comeback-clock, journey-lab
                                     never ran
```

#226 added an acceptance and shipped with **authoritative CI having never once executed it**. That is the worst thing a gate can do — not a wrong answer, but no answer wearing the same colour as one.

**Contamination.** The suites are not isolated (Coach Health reads `turn_ledger` database-wide, for one). On #224 I reported three product regressions from a batch run; each one evaporated when rerun after its own reset. A false diagnosis reached a lane decision before it was caught.

## What replaces it

One runner owning **reset → execute → collect → exit code**. The workflow's nineteen steps become two.

- **No `continue-on-error`.** A red is still a red gate; it just stops hiding what is behind it. The workflow already rejected yellow-for-red in writing for the Journey Lab, and that reasoning holds here.
- **No second PostgreSQL service, no reordering, no expectation weakened.** The order is byte-for-byte the order the YAML had.
- **The inventory moves into the runner, in one copy**, carrying each acceptance's reason for needing a real database rather than a fixture. Two owners of one inventory is how an acceptance gets added in one place and silently never runs in the other.

**The reset fails closed.** It refuses any non-loopback host, any non-PostgreSQL URL, `NODE_ENV=production`, and — outside CI — any run that did not explicitly opt in, so no hook or editor task can wipe a working database as a side effect. The host check is the decisive one: this product's real database lives behind a remote hostname, so it is out of reach by construction rather than by a name pattern someone could match by accident.

## The first full run is itself the evidence

```
PG ACCEPTANCE SUMMARY — 19 run, 17 green, 2 red
  RED    open-coaching-loop       exit 1
  RED    journey-lab              exit 1
pg-acceptance: RED — open-coaching-loop, journey-lab
```

Both reds are **pre-existing product failures in Codex's lane, untouched here**. The point is the second one: `journey-lab` is red on `2 · MULTI-DAY / CORRECTION — 33 checks, 1 failed`, and the old job **had never once reached it**. Isolating the acceptances did not create that failure, it revealed it — which is the whole argument for this cut.

## Controls — the runner graded against the real container

A test runner is the one thing its own suites cannot check: if it silently skipped everything, every suite would still "pass". So `script/pg-acceptance-runner-controls.ts` grades it with deliberate controlled reds, and runs **before** the runner in CI.

**19/19 green**, mapped to the issue's mandatory controls:

| # | Control | Check |
|---|---|---|
| 1 | early failure does not stop later ones | `every acceptance after a failure still executed` |
| 2 | job still exits RED | `the job still exits RED because one of them failed` |
| 3 | reset removes contamination | `a row seeded by one acceptance is gone before the next one starts` |
| 3′ | **its control** | `with the reset removed, the same pair contaminates and the second FAILS` |
| 4 | green set stays green | `nothing red, nothing invented — the gate passes` |
| 5 | unsafe DB refused | 7 refusals + **2 allow-controls** |
| 6 | scripts still standalone | `an individual acceptance still runs on its own, outside the runner` |

Control 3′ is the one that matters most: without it, "the row was gone" is equally well explained by the row never having been written. Control 5 is asserted **both ways** — a guard that refuses everything would satisfy all seven refusals and quietly stop the gate from ever running.

One more, not asked for but the same class of hole: **a reset that throws is recorded as a failure rather than run against whatever was left** — otherwise the contamination this removes walks back in through the error path.

## Runs

`tsc` clean · **37/37 suites** · **19/19 runner controls** · runner exit code 1 on the two product reds, 0 on a green set.

**No governor budget raised** — `modules` counts `server/` and `shared/` only, and neither is in the diff.

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -117,96 +117,23 @@ jobs:
       # so the forbidden command is not written here even in a comment.
       - name: Apply committed migrations to the ephemeral database
         run: npm run db:migrate
-      - name: Targeted multi-day food correction on real PostgreSQL
-        run: npx tsx script/pg-correction-acceptance.ts
-      - name: Safety early-return turn attribution on real PostgreSQL
-        run: npx tsx script/pg-safety-turn-acceptance.ts
-      # THE PROVENANCE OWNER IS A DATABASE FUNCTION (#184), so PostgreSQL is the only thing that
-      # can say whether its regular expressions compile. The broken version threw inside an
-      # AFTER INSERT trigger the application never awaits — a green suite, a normal reply, and
-      # every client-stated step count silently left untrusted.
-      - name: Step provenance bridge on real PostgreSQL
-        run: npx tsx script/pg-step-provenance-acceptance.ts
-      - name: Canonical client truth ordering on real PostgreSQL
-        run: npx tsx script/pg-client-truth-acceptance.ts
-      # TWO READERS, ONE SET OF FACTS (#179). Provenance, resolved_day and the workout ledger are
-      # COLUMNS, so only a real database can show that the canonical snapshot and the GPT
-      # fallback's context are reading the same rows — a fixture that answers every query the
-      # same way cannot tell two readers apart, which is how the divergence survived.
-      - name: Fallback reads canonical truth on real PostgreSQL
-        run: npx tsx script/pg-fallback-truth-acceptance.ts
-      # ONE COACH, PROACTIVELY AND REACTIVELY (#180). Held constraints are rows the client wrote
-      # today; whether the proactive decision reads them is only answerable against a database.
-      - name: Proactive decision authority on real PostgreSQL
-        run: npx tsx script/pg-proactive-authority-acceptance.ts
-      # A DAILY CONSTRAINT IS ABOUT WHAT SURVIVES (#194). The defect was a 24-message window, and
-      # a fixture that returns the same rows to every query cannot demonstrate a window.
-      - name: Daily constraint lifecycle on real PostgreSQL
-        run: npx tsx script/pg-daily-constraint-acceptance.ts
-      # THE CONSTRAINT REACHES EVERY MOUTH THAT NAMES A FOOD (#128). Which branch of the Next Meal
-      # menu a client reaches is composed from the day ledger — what is logged, what is left — so
-      # a fixture that answers every query the same way puts every client on one branch. That is
-      # how a menu with zero calls to the constraint owner looked fine.
-      - name: Food constraints reach the plate and grocery mouths on real PostgreSQL
-        run: npx tsx script/pg-food-constraint-mouths-acceptance.ts
-      # ONE ANSWER TO "WHICH WAY IS THE SCALE GOING" (#128). The verdict is computed from weigh-in
-      # ROWS — how many, how far apart, how recent — against an illness window in profile_notes,
-      # and the proactive half is graded on the message it actually sends, captured through the
-      # shadow door. Neither is expressible on a fixture that answers every query the same way.
-      - name: One weight-direction authority on real PostgreSQL
-        run: npx tsx script/pg-weight-authority-acceptance.ts
-      # ONE OPEN COACHING LOOP (#208). Restart durability, user isolation and exact SAST
-      # attribution require the real database; a fixture cannot prove compare-and-set closure.
-      - name: Open training move survives and resolves across turns
-        run: npx tsx script/pg-open-coaching-loop-acceptance.ts
-      # A SPARSE CLIENT IS COACHED, NOT JUST RECEIPTED (#203). Every decision here is computed
-      # from rows — how many days carry a meal, whether one carries TODAY, how stale the weigh-in
-      # is, and whether a same-day re-weigh updates or inserts. None of that is expressible on a
-      # fixture that answers every query the same way.
-      - name: Thin-evidence coaching on real PostgreSQL
-        run: npx tsx script/pg-thin-evidence-acceptance.ts
-      # INFORMATION VALUE (#213). Weekend coverage, stalled trend, the durable open loop and the
-      # later backdated answer are all real rows; this grades the one canonical INVESTIGATE owner.
-      - name: Highest-value investigation on real PostgreSQL
-        run: npx tsx script/pg-information-value-acceptance.ts
-      # ONE COACH SPEAKING (#207). The repetition defect this closes is invisible to a fixture
-      # that sends one message per client: it only appears across CONSECUTIVE durable writes by
-      # the same person, where the decision is recomputed from day state that the new event did
-      # not move. Real rows, real front door, several turns.
-      - name: Durable log turns are one coaching turn on real PostgreSQL
-        run: npx tsx script/pg-log-turn-acceptance.ts
-      # FACTS, RETRACTIONS AND SAME-TURN PROPAGATION (#211). Retraction-before-append ordering,
-      # one operation building on another's patch, and unrelated columns left alone are all
-      # properties of a real transaction — a fixture that returns whatever it was handed cannot
-      # show any of them, which is how a retracted diet came back inside its own write.
-      - name: Facts, retractions and same-turn propagation on real PostgreSQL
-        run: npx tsx script/pg-facts-acceptance.ts
-      # FOOD IDENTITY AND STATED PORTION (#206). Identity and portion are what get WRITTEN, and
-      # the invented-combo defects were visible only as kcal in the ledger — a matcher unit test
-      # cannot show that the row reaching meal_logs carries the right food at the right grams.
-      - name: Food identity and stated portion on real PostgreSQL
-        run: npx tsx script/pg-food-identity-acceptance.ts
-      # ONE WEIGHT AUTHORITY, ONE SPEAKABILITY VERDICT (#216). Illness contamination, window width
-      # and chronology are all properties of stored rows against stored dates, and the Monday
-      # claims only exist as a sent message — SHADOW=on captures them.
-      - name: Weight authority and speakability on real PostgreSQL
-        run: npx tsx script/pg-weight-speakability-acceptance.ts
-      # ONE RESTRICTION, EVERY SURFACE (#220). Consistency is a claim about what a CLIENT receives
-      # across surfaces reached by different handlers, different scheduler jobs and different cron
-      # minutes — a unit test on `allows` proves the predicate and nothing about who asked it.
-      # SHADOW=on captures the Sunday sends, so proactive food coaching is graded on the message.
-      - name: One restriction across every food surface on real PostgreSQL
-        run: npx tsx script/pg-restriction-consistency-acceptance.ts
-      # ONE COMEBACK CLOCK (#221). Both defects are claims about CHRONOLOGY across stored rows — a
-      # workout row's logged_at against a user's lastActiveAt, read through the real handler. A
-      # pure test can check the arithmetic; only this can show that the reply a returning client
-      # receives states the gap they actually experienced.
-      - name: One comeback clock on real PostgreSQL
-        run: npx tsx script/pg-comeback-clock-acceptance.ts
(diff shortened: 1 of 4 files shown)


ButiMM:
## Authoritative CI — the runner did the thing it exists to do

[Run 34231682066](https://github.com/ButiMM/KamLife-Coach/actions/runs/34231682066), head `0862bde`. `test` GREEN · `unit-baseline` GREEN · `pg-acceptance` RED, **by design and correctly**.

### The controls passed on the real container first

```
pg-acceptance-runner-controls: GREEN — all checks passed
```

That step runs **before** the runner and contains the deliberate controlled reds — a failing fixture that must not stop the ones behind it, a contamination pair that must fail with the reset removed, seven refused database configurations and two allowed ones. The runner is only trusted to grade the acceptances after it has itself been graded.

### Then all nineteen ran

```
PG ACCEPTANCE SUMMARY — 19 run, 17 green, 2 red
  GREEN  correction … weight-authority          (9)
  RED    open-coaching-loop       exit 1        <- the old job stopped HERE
  GREEN  thin-evidence, information-value, log-turn, facts,
         food-identity, weight-speakability,
         restriction-consistency, comeback-clock (8)
  RED    journey-lab              exit 1
pg-acceptance: RED — open-coaching-loop, journey-lab
##[error]Process completed with exit code 1
```

**Nine acceptances executed on authoritative CI that the previous run skipped entirely** — among them `restriction-consistency` (#220) and `comeback-clock` (#221), both of which had shipped without CI ever once running the proof they added.

### The second red is the point of the cut

`journey-lab` is red, and the old job **had never reached it**:

```
RED    2 · MULTI-DAY / CORRECTION — 33 check(s), 1 failed
       ✗ the day BEFORE the target is byte-identical
```

This is not a regression introduced here. It reproduces identically on `main` and matches the pre-existing failure I recorded on #224 and #226 from local runs. Isolation did not create it — it **revealed** it, which is the entire argument for the cut. Same for `open-coaching-loop`: same two check names as on base, Codex's seam, untouched by this PR.

### On the red gate

Per the issue: *"Do not use the current `pg-open-coaching-loop` product red as an excuse to stop the runner."* I did not — the job runs to completion and records both reds, and it still exits non-zero, so the gate is intact. **This PR's `pg-acceptance` red is two product failures in Codex's lane, not an infrastructure failure**, and fixing them is out of this cut's scope and inside a lane I am told not to touch. There is nothing here for me to push: the check is red because two product acceptances are red, which is exactly what a working gate should say.

Both failures are now *visible* on every future run instead of one hiding the other — which is the difference this PR was asked to make.

### No re-run spent

A re-run would tell us nothing: these are deterministic product assertions, not flake, and both were measured locally before the push.

---

ButiMM:
## CTO GATE — THREE INFRASTRUCTURE BLOCKERS FROM CURRENT REVIEW

Stay on #227. Do **not** touch product/runtime logic. The runner concept is correct and the first full run proved the value, but three review findings are real and directly violate the cut's safety/observability contract:

1. **P1 — acceptances must still run if the controls step itself fails.** GitHub's default success condition currently means a controls failure can skip the runner and recreate early-exit blindness. Make the runner execute on non-cancelled prior failure while preserving the overall job RED. Do not use blanket `continue-on-error`.
2. **P1 — the controls script must guard before any destructive DB operation.** It currently accepts any non-empty `DATABASE_URL` before importing/using the reset path. Apply the same fail-closed safety check before pool import/reset/truncate so a direct/local invocation cannot touch a remote or unsafe database.
3. **P2 — table TRUNCATE is not full isolation when an acceptance mutates schema/functions.** `pg-step-provenance-acceptance.ts` can replace a PostgreSQL function and only restore it near the end; a crash can leave later suites on mutated schema. The runner must guarantee clean schema-level state between acceptances (or an equivalent safe restoration mechanism), not only clean rows.

### Proof required
Keep one PostgreSQL service/job and one inventory. Add/extend controls so each of the three findings is independently demonstrated. The decisive proof must show: controls can fail and all acceptances still run; job stays RED; unsafe direct invocation refuses before destructive work; schema mutation from one failing acceptance cannot contaminate the next; fully green path remains green.

Then rerun authoritative CI on newest main and return exactly:

`FIXED + PROVEN — READY FOR CTO — BUILDER STOPPED`

with head, CI, and clean-state packet. No next task and no product changes.

ButiMM:
## Three Codex findings closed — `0862bde` → `f4f2e49`

All three are holes in **this cut's own promise**, not in the product. Each reproduced before it was fixed.

### P1 — a red controls step *skipped the runner*

The sharpest of the three, and the most embarrassing: a GitHub Actions step runs by default only if every step before it succeeded. So the step that grades the runner could take all nineteen verdicts down with it — **the exact early-exit blindness this cut removes, rebuilt at the seam the cut itself added.**

Not hypothetical, either. The controls executed a real acceptance to prove standalone runnability, so a genuine product regression there would have silenced the entire acceptance run.

```yaml
- name: Every acceptance, isolated, on real PostgreSQL
  if: ${{ !cancelled() }}
```

Not `continue-on-error` — both steps still fail the job. It says only that the evidence is worth collecting even when the collector's grader is unhappy.

The coupling underneath was worth removing too: control 6 asserted the standalone acceptance came back **GREEN**, which made an infrastructure control depend on product health. What requirement 6 is actually about is whether the script still *runs* on its own, so it now asserts it reached its own verdict line — GREEN or RED.

### P1 — the controls script reset before checking where it pointed

It tested `testDatabaseSafety` as a pure function while never applying it to its own destructive work. A non-empty `DATABASE_URL` was the only thing between the command and truncating a real database — the precise hazard this cut exists to close, sitting in the file whose job is to prove it closed. The guard now runs **first**, before the pool is imported, and a database that is present but not a throwaway is a **REFUSAL** (exit 2), not a skip.

### P2 — truncation is not a clean database

The finding I most wanted to be wrong, and it isn't. Acceptances mutate **schema**, not only rows:

```
pg-step-provenance-acceptance  installs the pre-#184 faulty kamlife_parse_step_report
                               …3 statements…
                               restores the working one
```

An exception in that window leaves the broken parser installed, where `TRUNCATE` cannot reach it. Every later acceptance — Journey Lab especially — would then fail for reasons belonging to nothing in the diff. My reset promised "the same clean database" and did not deliver it.

The reset now rebuilds: drop the schema, drop the ledger, re-apply the committed migrations.

**Both schemas, and that is not a detail.** Drizzle keeps `__drizzle_migrations` in its own `drizzle` schema, so dropping `public` alone leaves it claiming every migration is applied and the re-migrate silently does nothing — a rebuild that reports success and produces an *empty* database. I found that by measurement: dropping `public` only left **0 tables**.

### Controls now 24/24, each new one carrying its opposite

```
=== 7 · THE RESET RESTORES SCHEMA AN ACCEPTANCE BROKE, NOT JUST ROWS ===
  PASS  the migrated schema carries the step-provenance function to begin with
  PASS  CONTROL: an acceptance really can leave a mutated function behind
  PASS  …and the reset puts the committed definition back, byte for byte

=== 8 · THESE CONTROLS REFUSE AN UNSAFE DATABASE BEFORE TOUCHING IT ===
  PASS  pointed at a remote database, this script refuses and resets nothing
```

Control 7's middle line is the one that matters: without it, "the function was correct after the reset" is equally well explained by it never having been broken.

### Effect on the run

**Unchanged: 19 executed, 17 green, 2 red, in two minutes.** The stronger reset costs ~1.5s per acceptance and moved no verdict. `tsc` clean, 37/37 suites, no governor budget raised.

The two reds remain `open-coaching-loop` and `journey-lab` — both pre-existing product failures in Codex's lane, untouched here, and both now visible on every run instead of one hiding the other.

---

ButiMM:
## All three gate blockers closed and independently demonstrated — head `abe1f1c`

Each of the three is fixed **and** proven by a control that carries its own opposite, so none of them can pass by accident.

### Blocker 1 — acceptances must run even if the controls step fails

Fixed with `if: ${{ !cancelled() }}` on the runner step. No blanket `continue-on-error`; both steps still fail the job.

**This one could not be proven locally** — it is a property of GitHub's step condition, not of any script. So I pushed a commit that forced the controls red for exactly one run, captured the evidence, and reverted it in the very next commit.

[Run 34243099239 on `f159f30`](https://github.com/ButiMM/KamLife-Coach/actions/runs/34243099239):

```
  FAIL  TEMPORARY PROOF: a deliberately failing control, reverted in the next commit
pg-acceptance-runner-controls: RED — 1 check(s) failed

…and the very next step still ran:
PG ACCEPTANCE SUMMARY — 19 run, 17 green, 2 red
```

Controls red, all nineteen verdicts still collected, job still red. `f159f30` is reverted by `abe1f1c`; `git diff a7b14fc abe1f1c` is **empty**.

### Blocker 2 — guard before any destructive operation

`testDatabaseSafety` now runs at the top of the controls script, **before the pool is imported**. A missing `DATABASE_URL` is a SKIP; a database that is present but not a throwaway is a **REFUSAL** with exit 2, which cannot read as a pass.

```
=== 8 · THESE CONTROLS REFUSE AN UNSAFE DATABASE BEFORE TOUCHING IT ===
  PASS  pointed at a remote database, this script refuses and resets nothing
```

That control spawns this very script against a remote host and asserts it refuses — the guard proven *gating*, not merely proven correct in isolation.

### Blocker 3 — schema-level isolation, not just rows

The reset now rebuilds: drop `public`, drop the `drizzle` ledger, re-apply committed migrations. Both schemas matter — dropping `public` alone leaves the ledger claiming everything is applied and produces an **empty** database while reporting success (measured: 0 tables).

Graded exactly as the gate words it — a mutation left by a **failing** acceptance:

```
=== 7 · THE RESET RESTORES SCHEMA AN ACCEPTANCE BROKE, NOT JUST ROWS ===
  PASS  CONTROL: the mutating acceptance really did fail
  PASS  a schema mutation left behind by a FAILING acceptance cannot reach the next one
  PASS  CONTROL: with row-only truncation the broken parser survives and the next acceptance FAILS
  PASS  …and the rebuild restores the committed definition byte for byte
```

The third line is the decisive one: it reruns the identical pair under the old row-only truncation and requires the second acceptance to **fail**. Without it, "the parser was fine afterwards" is equally well explained by a mutation that never landed.

### Authoritative CI on the clean head — [run 34243580702](https://github.com/ButiMM/KamLife-Coach/actions/runs/34243580702)

`test` GREEN · `unit-baseline` GREEN · `pg-acceptance` RED, correctly:

```
pg-acceptance-runner-controls: GREEN — all checks passed     (26/26)
PG ACCEPTANCE SUMMARY — 19 run, 17 green, 2 red
  RED  open-coaching-loop   FAIL  the production reply actually delivers the canonical move
                            FAIL  the reactive move survives after the originating turn returns
  RED  journey-lab          ✗ the day BEFORE the target is byte-identical
```

Both reds are pre-existing product failures in Codex's lane, untouched here. One PostgreSQL service, one job, one inventory, order unchanged, no product or runtime code in the diff.

---

ButiMM (commented):
CTO adjudication complete. Independent lane: #228 passed the final infrastructure gate on head abe1f1c and was merged as 9dc6f289. The red pg-acceptance status is the intended reporting of pre-existing product failures, not a #227 infrastructure regression. No product/runtime code was changed.

PR #225, opened 2026-09-08
Make behavioural patterns evidence-backed (#217)

ButiMM:
Closes #217.

## First divergence

The existing durable profile owner, `client_intelligence_profiles.pattern_flags`, stored lifetime logging/silence heuristics as prose-like flags. It had no attributable outcome provenance, no bounded support strength, no recency/decay, and no explicit path into the canonical decision owner. Missing logs could therefore be mistaken for behaviour while real repeated outcomes could not affect the next decision.

## Bounded correction

- Extend the existing profile JSONB owner with versioned, evidence-backed training patterns.
- Source failure/success only from attributable #208 open-loop outcome rows and completed workout rows.
- Require two distinct weeks before a negative pattern becomes active.
- Decay after 42 days and supersede after two recent contradictory completion weeks.
- Carry support/confidence, evidence row IDs/days/sources, and intervention provenance.
- Feed only structured pattern context into the existing `chooseAction` owner across reactive, proactive, command, and morning paths.
- Preserve #208 open-loop, #213 INVESTIGATE, safety/illness, held-constraint, and doNotMention precedence.
- Record whether the delivered canonical training move was the existing minimum-training intervention so later outcomes can be attributed.
- Add real-PostgreSQL acceptance to the authoritative PG job.

No new memory store, parser family, decision ladder, response author, migration, or model-authored pattern truth.

## Local proof

- `npm run check` — PASS
- `npm run test:continuity` — PASS
- unit baseline comparator — PASS: base 1058/1062 (4 known), head 1059/1063 (same 4), NEW 0
- production parity — 141/142, same existing weekly-owner harness assertion as main
- schema safety — PASS
- Journey Lab — correctly reports SKIPPED locally without DATABASE_URL
- Windows `npm test` runner cannot spawn its extensionless Linux `node_modules/.bin/tsx` path and reports all suites as instant timeouts; authoritative Ubuntu Actions is the grading environment.
- Current-main structural governors retain their existing merged-main reds locally; no budget was raised.

The PR Actions PG container is the authoritative proof for longitudinal persistence/order, #208 open-loop continuity, Journey Lab, and all #217 controls.

PR #231, opened 2026-09-08
#221 journey 4: name what the recap counts, so it stops contradicting itself

ButiMM:
Journey 4 of #221. **Do not merge — at the CTO gate.** BASE `main@5da8e5f` (one rebase onto the merged #225, no rediscovery of journeys 1–3).

## The defect, reproduced through the real front door before anything changed

A client with four workout rows in the last seven days and no meal rows:

```
*Zanele's Progress*

🔥 Today: *nothing logged yet*
💪 Sessions this week: *4*
📋 Days logged (7d): *0/7*
```

Two consecutive lines, one owner, and they disagree. A client who trained four times is told they logged nothing.

`window.daysLogged` counts **days carrying a meal** — `day-ledger-core` says exactly that in its own definition, and it is the divisor for the averages on the next line. **The number was right. The word was wrong.** Shown unqualified as "Days logged" it reads as a claim about everything the client did, one line under the training count it silently excludes.

Four customer-visible mouths carried it, three sitting directly beside a session count — both progress cards and all three Sunday check-in variants, where **"4 sessions, 0 days logged"** went out as one sentence.

## The fix

Name the quantity for what it counts, at the mouths that state it:

```
💪 Sessions this week: *4*
🍽️ Food logged: *0/7 days*
```

## The value is untouched, and that is the point

The tempting fix is to widen `daysLogged` to count training days too. It divides `avgKcal` and `avgProtein` and **#203's thin-evidence coaching keys off it**, so widening it would move coaching decisions and every average on every card. Two controls assert the count stays put:

```
PASS  CONTROL: three days carry food, so the count is 3/7 and not 5/7
PASS  CONTROL: …and the session count is still its own number
```

**Red on revert, three mechanisms, all three red** — including that widening as the opposite defect:

```
── the Progress card says 'Days logged' again                     RED — 3
── the 7-day breakdown says 'Days logged' again                   RED — 1
── the food-day COUNT is widened to include training (opposite)   RED — 3
```

The third one goes red on the *value* controls rather than the label ones, which is the proof they are testing different things.

## What did NOT reproduce — recorded, not quietly dropped

**The contradictory session numbers across surfaces (12 / 6 / 13) do not reproduce on current main.** I tried to make the lifetime counter drift from the ledger through ordinary use and could not:

| route | result |
|---|---|
| log the same day twice | guarded — `already logged`, rows and counter both stay |
| repeat a backdated day | guarded — `already got yesterday's workout logged` |
| named-day backdate | correct — today is Tuesday, so "Monday" *is* yesterday |

I nearly filed that last one as a defect before checking the calendar. It is not one.

What remains true: `users.totalWorkoutsCompleted` is a second answer to a question `workout_logs` already answers, and **nothing reconciles them**. Its increment paths are guarded today, but if it ever diverges — as it evidently did on the real client behind the banked 13-vs-7 evidence — three surfaces repeat the wrong number forever with no path back. That is a real gap. It is not the reported defect and is **not fixed here**.

### One design decision, which is yours and not mine

If the counter and the ledger disagree, which is truth? Deriving from `workout_logs` would tell a legacy client they have done **fewer** sessions than they were told yesterday — a visible regression on their own history. Taking `max` never reduces anyone's count and makes future drift impossible, but preserves whatever legacy inflation already exists. I did not pick one silently.

Also recorded: `"I trained Monday and Tuesday"` attributes **one** day. The multi-day attribution owner exists and is deliberately unwired — declared debt in the `unreachableCapabilities` budget, pre-existing, out of this cut's bounds.

## Runs

`tsc` clean · **37/37 suites** · **21/21 PG acceptances GREEN**, each isolated by the #227 runner:

```
PG ACCEPTANCE SUMMARY — 21 run, 21 green, 0 red
pg-acceptance: GREEN — every acceptance executed and passed
```

Worth noting: **`open-coaching-loop` and `journey-lab` are both green now.** They were the two standing product reds; #225 fixed them. This is the first fully green PG gate since the runner landed.

New acceptance wired into the runner inventory. No governor budget raised. No new owner, no second counter, no new response author.

---

### DIFF
--- a/script/pg-acceptance-runner.ts
+++ b/script/pg-acceptance-runner.ts
@@ -150,6 +150,13 @@ export const ACCEPTANCES: Acceptance[] = [
     // workout row's logged_at against a user's lastActiveAt, read through the real handler.
     command: ["npx", "tsx", "script/pg-comeback-clock-acceptance.ts"] },
 
+  { id: "session-recap", title: "One training recap that does not contradict itself",
+    // ONE RECAP (#221 journey 4). The contradiction only exists across two row families — workout
+    // rows present, meal rows absent — composed into one card by one owner. A fixture that answers
+    // every query the same way cannot put a client in the state where the card disagrees with
+    // itself, which is how "Sessions this week: 4" sat above "Days logged (7d): 0/7".
+    command: ["npx", "tsx", "script/pg-session-recap-acceptance.ts"] },
+
   { id: "journey-lab", title: "Six critical journeys through the real system",
     // THE SIX JOURNEYS (#170). Same database, same migrations, same front door — a second job
     // would be a second copy of this infrastructure for no gain. It is in this runner for the same
--- a/script/pg-session-progression-trace.ts
+++ b/script/pg-session-progression-trace.ts
@@ -1,0 +1,84 @@
+/**
+ * TRACE — #221 journeys 4–7, session/progression truth on real PostgreSQL.
+ *
+ * Read-only reproduction, run before anything changes. One client, one sitting, the surfaces that
+ * each state a session number, against the durable rows those numbers claim to describe.
+ */
+process.env.OPENAI_API_KEY = "sk-test-offline";
+process.env.OFFLINE_AI = "1";
+process.env.NORMALIZER = "off";
+process.env.ENGINE_LIVE = "off";
+process.env.PROACTIVE_PAUSED = "true";
+process.env.TWILIO_ACCOUNT_SID = "ACtest00000000000000000000000000";
+process.env.TWILIO_AUTH_TOKEN = "test";
+process.env.TWILIO_WHATSAPP_NUMBER = "+27000000000";
+process.env.NODE_ENV = "production";
+
+const R = console.log.bind(console);
+const { pool, db } = await import("../server/db.ts");
+const schema = await import("../shared/schema.ts");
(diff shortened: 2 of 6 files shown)


ButiMM:
## Authoritative CI — fully green, and every acceptance actually ran

[Run 34251357000](https://github.com/ButiMM/KamLife-Coach/actions/runs/34251357000), head `9e66911`. `test` GREEN · `unit-baseline` GREEN · `pg-acceptance` GREEN.

Read from the log rather than the badge — the runner controls pass first, then all twenty-one execute:

```
pg-acceptance-runner-controls: GREEN — all checks passed
PG ACCEPTANCE SUMMARY — 21 run, 21 green, 0 red
```

**This is the first fully green PG gate since the #227 runner landed.** `open-coaching-loop` and `journey-lab` — the two standing product reds this branch inherited — are both green on merged main; #225 closed them. Worth stating plainly because it cuts the other way from the last three reports: the runner is now confirming a clean board, having spent its first two runs exposing a red one.

No review comments outstanding. Green and mergeable, holding at the gate.

The one open question from the PR body stands and is yours, not mine: when `users.totalWorkoutsCompleted` and the `workout_logs` ledger disagree, which is truth? Deriving tells a legacy client they have done fewer sessions than they were told yesterday; `max` never reduces anyone but preserves whatever inflation already exists. Journey 5 waits on that answer.

---

PR #232, opened 2026-09-08
#221 journeys 5–7: lifetime history is not a programme position

ButiMM:
Journeys 5–7 of #221, to the adjudicated rule. **Do not merge — at the CTO gate.** BASE `main@f399309`.

## The defect

A legacy client — twelve on the lifetime counter, seven durable workout rows — opened today's session and read:

```
*Foundation Phase · Week 4 · Session 13*
*Day 2 — Today's Workout*
```

Two numbers, two clocks, and one of them answering a question it cannot answer. **"Session 13" is a claim about where this client is in the programme**, built from `totalWorkoutsCompleted` — which is lifetime history, kept precisely because clients who started before every session had a durable row may own sessions no row can prove. The line directly beneath it already states the real position, from the owner that holds it.

## The fix, and the two shortcuts not taken

Per the adjudication — **not** `workout_logs.length` (erases legitimate history), **not** `max(counter, rows)` (promotes drift into programme truth). The counter stays exactly what it is, and says so:

```
*Foundation Phase · Week 4 · 12 sessions overall*
```

That is the convention `sessionHeaderLine` already adopted after the same defect reached a client in July — `Session N overall` — applied to the one header left behind.

## Proof — the exact ugly state you named

`rows = 7`, `counter = 12`, cursor at week 4 day 2, run against five surfaces. **24/24 green.**

```
=== 5 · FIVE SURFACES, ONE PROGRESSION, DESPITE A LEGACY MISMATCH ===
  PASS  the ugly state is real: 7 durable rows against a lifetime counter of 12
  PASS  today's session header does not claim a programme position from the counter
  PASS  the next session does not either
  PASS  …and neither does the programme view
  PASS  today's workout is day 2 — the cursor's day
  PASS  …and the next session names the same day 2
  PASS  CONTROL: asking three questions moved neither the cursor nor the counter
  PASS  completing a session writes one durable row
  PASS  …and advances the programme cursor, which is what owns 'which session is next'
  PASS  …and moves lifetime history by exactly one, still ahead of the rows and still legacy
  PASS  the next session follows the cursor, not the counter
```

### Journeys 6 and 7 needed no change — now pinned rather than assumed

```
=== 6 · BACKDATED TRAINING UPDATES THAT SAME TRUTH, AND ONLY IT ===
  PASS  a backdated session writes its durable row
  PASS  …and moves lifetime history by one
  PASS  …and does NOT move the programme cursor — a past day answers nothing about today
  PASS  CONTROL: reporting the same day twice adds neither a row nor a lifetime session

=== 7 · A QUIET DAY OR AN UNRELATED MESSAGE FABRICATES NOTHING ===
  PASS  one quiet day with comeback words is not a comeback
  PASS  an unrelated question states no programme position
  PASS  …and neither message moved the cursor, the counter or the ledger
```

I found both already correct when I probed them, and said so rather than manufacturing a fix. What they lacked was a test that would notice if that stopped being true — P0-3's cursor discipline in particular was protected only by a comment.

## Red on revert — three mechanisms, all three red

```
── the header claims a programme position from the counter        RED — 3
── the lifetime number is DELETED rather than relabelled          RED — 2   (opposite defect)
── a backfill advances the programme cursor again                 RED — 2
```

The middle one is the one worth reading. **Deleting the number satisfies every position check** — no number, no position claim — and silently loses the legacy history the counter exists to carry. Section 5b is the only thing standing between this cut and that "fix".

The third restores the exact P0-3 defect: "I trained on Monday" advancing today's cursor, so the client is correctly told Monday was logged and silently loses today's session slot.

## Runs

`tsc` clean · **37/37 suites** · **22/22 PG acceptances GREEN**, each isolated by the #227 runner. New acceptance wired into the inventory. No governor budget raised. No new owner, no second counter, no migration, no new response author.

---

### DIFF
--- a/script/pg-acceptance-runner.ts
+++ b/script/pg-acceptance-runner.ts
@@ -157,6 +157,14 @@ export const ACCEPTANCES: Acceptance[] = [
     // itself, which is how "Sessions this week: 4" sat above "Days logged (7d): 0/7".
     command: ["npx", "tsx", "script/pg-session-recap-acceptance.ts"] },
 
+  { id: "session-owner", title: "One progression owner, despite a legacy lifetime mismatch",
+    // ONE PROGRESSION OWNER (#221 journeys 5-7). The whole proof is a legacy client whose lifetime
+    // counter disagrees with the durable ledger — seven rows against twelve — read through five
+    // different handlers that must still agree about where the client is in the programme. A
+    // fixture that answers every query the same way cannot produce that disagreement, and the
+    // cursor discipline on a backfill is a claim about what a write did NOT touch.
+    command: ["npx", "tsx", "script/pg-session-owner-acceptance.ts"] },
+
   { id: "journey-lab", title: "Six critical journeys through the real system",
     // THE SIX JOURNEYS (#170). Same database, same migrations, same front door — a second job
     // would be a second copy of this infrastructure for no gain. It is in this runner for the same
--- a/script/pg-session-owner-acceptance.ts
+++ b/script/pg-session-owner-acceptance.ts
@@ -1,0 +1,230 @@
+/**
+ * REAL-POSTGRESQL ACCEPTANCE — one progression owner (#221, journeys 5-7).
+ *
+ * THE RULE THIS ENFORCES, as adjudicated:
+ *
+ *   Current session, next session, programme week/day and backdated progression come from the ONE
+ *   existing progression owner — the programme cursor (`programmeWeek` / `programmeDayInWeek`) and
+ *   what the durable ledger says was done. NEVER from the lifetime counter.
+ *
+ *   `users.totalWorkoutsCompleted` survives as LEGACY LIFETIME HISTORY, because clients who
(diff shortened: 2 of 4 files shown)


ButiMM:
## CTO NEXT QUEUE — BUILDER ROLE LOCKED

Finish/adjudicate this PR only. Do **not** perform a broad convergence sweep after merge.

Claude remains a builder. CTO + Grok own diagnosis/review. After this PR is merged, CTO will give Claude the next bounded implementation task from live tester / Coach Health evidence with the first divergence and existing owner already identified.

No copy/paste handoff is needed; the next build instruction will be posted directly to its GitHub issue.

ButiMM:
## CTO NEXT BUILD — START NOW

#232 is merged and deployed on Railway at `main@a15166a5`.

Your next bounded build is **#234 — LIVE P1: normalizer must not invent meal slots or dates**. The production reproduction, first divergence, truth rule, owner, controls and stop condition are already in that issue. Do not sweep or rediscover the system; start from #234 and execute it.

No copy/paste handoff required.

PR #244, opened 2026-09-12
A long voice note reaches the handlers whole

ButiMM:
Cut 3. Base `ca1a85c161b8e0d796dc23e9b24cb783fa1f7329` (fresh main, #245 merged). Head `95221747f5ba11f22c9e5eb74ffcd440d70d7719`.

**A long voice note reaches the handlers whole, and the cleaner may repair spelling without rewriting clauses.**

## The four silent cuts, measured on `017efd9`

A 2,408-character note — 500 words, about three minutes of speech:

```
handed to the model            1500 chars
returned as "the transcript"   1500 chars
SILENTLY DELETED                908 chars
```

| fact | where | survived on `017efd9` |
|---|---|---|
| named-day meal | beginning | kept |
| 8,500 steps | middle | kept |
| the workout correction | near the end | **DELETED** |
| the feeling | throughout | kept |
| question 1 | near the end | **DELETED** |
| question 2 | near the end | **DELETED** |
| the final tail marker | last words | **DELETED** |

| # | where | what it cost |
|---|---|---|
| 1 | the cleaner's 1,500-char window | 908 chars of a three-minute note |
| 2 | a ceiling with **no floor** | a reply cut off by `max_tokens` became the transcript |
| 3 | the condenser's 4,000-char window | reachable, and it replaced the account with a retelling |
| 4 | **the ledger's own 2,000-char cap** | the durable record, cut by 408 |

The fourth was found by this cut's acceptance, not by reading code.

## The cleaner's contract — four reviews, five holes, one rule

Two percentages were tried and **both were beaten**:

| review | reply | why it passed |
|---|---|---|
| first | 60% of the head, `finish_reason` either way | the 50% floor and a 40% word-overlap test accepted it |
| second | **95.84%** of the head, ending intact, `stop` | a length floor cannot see a hole in the middle; `retainsOriginal` compared two **sets**, and a set has no order |

Deleted in the second case: `"Actually I missed my Tuesday workout. What should I do today?"`

Then the ordered edit contract that replaced them was itself reviewed, and **two holes were reproduced in it** at head `70ed20c`. Both are fixed here, in the same predicate, at `9522174`:

| hole | reproduced | now |
|---|---|---|
| **a near-neighbour of a food word counted as a spelling repair** — membership in an SA vocabulary plus edit distance ≤ 3 | `"I have pain"` → `"I have pap"`, `"I am sad"` → `"I am pap"` | only an **explicitly vetted `FROM→TO` pair** is allowed. `VETTED_REPAIRS` holds one entry, `stamp → samp`. The distance metric and the target vocabulary are **deleted, not tuned** |
| **the tokenizer ate the sign and the decimal point**, and discarded non-ASCII lexical content entirely | `"my change was -5 kg"` → `"my change was 5 kg"`; `"8.5"` → `"85"` | a quantity is **one token carrying its leading sign and its internal separators**; everything else is a run of Unicode letters, marks and digits |

A client reporting pain or low mood had it recorded as a plate of food. Five kilograms lost read back as five gained. Neither was visible to any handler.

**A share of the text cannot distinguish a spelling from a sentence**, and **a distance metric cannot distinguish a repair from a guess.** The rule is now:

- the cleaned head must be the **same tokens, same order, same repetition**
- case, spacing and punctuation **between tokens** are free — punctuation is not *globally* free: a sign and a decimal point are part of the number
- a token may be replaced only when **that exact pair is listed** in `VETTED_REPAIRS`
- **numbers, weekdays and negations may never be substituted at all**
- anything else returns the **whole raw transcript**

Growing the pair map is a deliberate act with a name attached; guessing from a metric is not, which is why the metric is gone rather than retuned. This errs strict on purpose — `"8500"` and `"8,500"` are different tokens here and a reply that regroups digits is refused. **Refusing returns the client's own raw words, which is the safe half.**

The length floor and the ending check are **deleted, not loosened**: the contract subsumes both, and two gates that can no longer fail on their own is how a suite starts grading nothing while looking thorough.

## What changed

| file | change |
|---|---|
| `sa-transcript.ts` | `splitForClean` sends one window and **rejoins the rest untouched**; `onlyApprovedRepairs` replaces the set-overlap check; `VETTED_REPAIRS` replaces vocabulary-plus-distance; `lexicalTokens` preserves signs, decimals and non-ASCII letters; `finish_reason` must be `stop`; `condenseVoiceRamble` deleted |
| `media.ts` | `forBrain` **is** the cleaned transcript, unconditionally |
| `chat-log.ts` | the ledger records the whole input; a cut says so *inside the value* |
| `utils.ts` | `transcriptMustPassWhole` / `isMessyLifeTranscript` / `transcriptIsLogList` deleted with the gate they served |

**The condenser could not be kept as an aid.** There is one routed string; a condensation offered "alongside" the transcript is a second version of what the client said.

**What that costs:** a three-minute note now reaches the coaching model in full.

## Evidence

`script/pg-long-voice-tail-acceptance.ts` — real PostgreSQL, real front door, graded on `turn_ledger` and the durable owners. §2 proves four of the seven facts live beyond the old window, so §1 grades the fix and not the fixture.

Graded in `script/voice-provenance-tests.ts`, every case driven through the **real** `cleanSATranscript` against a stub network boundary:

```
the exact reported case   a clause deleted from the middle, 95.84% kept
a negation inverted       "I missed my session" -> "I finished my session"
a quantity rewritten      8500 -> 8000
a named day rewritten     Tuesday -> Thursday
an invented sentence      a line the client never said

THE FOUR-REVIEW CASES, reproduced exactly as reported:
"I have pain"             -> "I have pap"           REFUSED
"I am sad"                -> "I am pap"             REFUSED
"I ate porridge"          -> "I ate pap"            REFUSED  (unvetted, though pap is real)
"my change was -5 kg"     -> "5 kg"                 REFUSED
"my change was −5 kg"     -> "5 kg"                 REFUSED  (unicode minus)
"I am 8.5 kg down"        -> "85 kg"                REFUSED
"I am 8.5 kg down"        -> "8,500 kg"             REFUSED
a non-ASCII word deleted                            REFUSED
a non-ASCII word invented                           REFUSED

AND THE POSITIVES, so a cleaner that refuses everything cannot pass:
stamp -> samp             still accepted, as it must be
punctuation and case      free
a decimal left alone      still accepts a punctuation repair
an untouched non-ASCII note  still accepts a punctuation repair
an unchanged transcript   passes through untouched
```

The non-ASCII case carries **the superseded tokenizer as a local control**, asserting the deleted word left *no trace at all* under it — so the fixture is not vacuous and the check is not grading something the old code would also have caught.

**red-on-revert 11/11**, both unmodified controls passing first:

```
CONTROL  both graders GREEN unmodified — detections below are real
1   the window truncates again          voice-provenance-tests
2   THE EDIT CONTRACT REMOVED           acceptance + voice-provenance-tests
2b  the finish-reason gate removed      voice-provenance-tests
2c  protected tokens unprotected        acceptance + voice-provenance-tests
2d  THE VOCABULARY COMES BACK           acceptance + voice-provenance-tests
2e  THE ASCII-ONLY TOKENIZER COMES BACK acceptance + voice-provenance-tests
3   the condenser returns               both
4   the ledger cap returns              acceptance
5   the split drops the tail            voice-provenance-tests
6   CONTROL: the cleaner stops cleaning voice-provenance-tests
7   CONTROL: the record mangles numbers acceptance
```

2d and 2e turn red **on behaviour, not only on a source string** — under 2d the `pain` and `low mood` cases fail as executed assertions.

A detection is a **graded failed assertion**: the grader must reach its verdict line, that verdict must report failures, and a `FAIL` must be printed. A crash fails the harness even when the other grader legitimately went red.

## Outstanding, under Cut 6 — not absorbed here

The long note still does not log the meal it names, and the reply addresses neither question. Measured final body:

```
Lerato — 8,000 steps is your target. No steps logged yet today — send your count:
"8,500 steps" or "walked 5km".
```

**That sentence is true.** Their 8,500 is stored against **yesterday**, which is what they said; "no steps logged yet today" is not a false claim and is not described as one. What is wrong is that the turn misses the food and answers neither question — first-match-wins gives one turn to one owner. §5 of the acceptance records it and asserts nothing either way.

Cut 3 makes the client's account arrive whole. Cut 6 must make the coach act on it.

## Gates on `9522174`

```
tsc --noEmit          clean
npm test              38/38 green, none skipped
unit-tests            1062/1062
pg-long-voice-tail    GREEN
red-on-revert         11/11, with both controls
governors             at budget, none raised
regexLiterals         442/442 UNCHANGED — the new tokenizer stays INLINE, as the
                      pattern it replaces was. It is one predicate's tokenizer,
                      not a named pattern authority that decides anything about
                      a message; naming it would have moved the counter for a
                      rewrite that adds no such authority.
```

## Traps walked into, recorded

A doc block explaining `retainsOriginal`'s removal matched an assertion searching for it, reporting the code was back — comments are stripped before that check now. A `finish_reason` fixture whose reply equalled its input produced the same string whether accepted or refused, so it passed with the gate deleted. And the first non-ASCII fixture removed a neighbouring ASCII word along with the non-ASCII one, which would have made the deletion visible for the wrong reason — its control caught that before the case was trusted.

### DIFF
--- a/script/check-architecture.ts
+++ b/script/check-architecture.ts
@@ -62,7 +62,7 @@ const BUDGET = {
    * to the true figure in one deliberate commit. Until then this red line is the marker, and it is
    * the ONLY thing in this guard that is red — one red line means something, four never did.
    */
-  regexLiterals: 446,
+  regexLiterals: 442,
   /**
    * GUARD #13 — see unreachableExports above. Sixty-two capabilities cannot be reached by a
    * client message today. This budget is deliberately set THREE BELOW that, so this guard is RED
@@ -273,7 +273,7 @@ const RAISES: Array<{ key: keyof typeof BUDGET; from: number; to: number; date: 
       + "so 21 is the smallest truthful current baseline. FROM HERE IT FALLS ONLY.",
   },
   {
-    key: "regexLiterals", from: 318, to: 446, date: "2026-08-24 (fell to 448 on 2026-09-05, to 447 on 2026-09-07, to 446 on 2026-09-09)",
+    key: "regexLiterals", from: 318, to: 442, date: "2026-08-24 (fell to 448 on 2026-09-05, to 447 on 2026-09-07, to 446 on 2026-09-09, to 442 on 2026-09-12)",
     why: "NOT A RAISE — A CORRECTED MEASUREMENT, and the follow-up this budget's own comment "
       + "declared owed on 2026-08-17: \"repair the matcher to see multi-line assignments and "
       + "re-baseline to the true figure in one deliberate commit.\" This is that commit. The "
@@ -299,7 +299,13 @@ const RAISES: Array<{ key: keyof typeof BUDGET; from: number; to: number; date: 
       + "let buildPatternSummary infer training from chat wording — is deleted; the weekly training "
       + "signal now reads workout_logs, the durable owner that same function already queried for "
       + "its 28-day count. One pattern fewer because a SECOND ANSWER to \"did they train\" was "
-      + "removed, not because anything unrelated was compressed to make room.",
+      + "removed, not because anything unrelated was compressed to make room. AND IT FELL AGAIN: "
+      + "446 -> 442 on 2026-09-12 (Cut 3). Four patterns left with server/utils.ts "
+      + "transcriptMustPassWhole, isMessyLifeTranscript and transcriptIsLogList, deleted whole. They "
+      + "were the gate on the voice summariser — WHICH notes may be shortened — and the summariser is "
+      + "gone, so the only answer they could still give was no. A predicate with one possible answer "
+      + "is not a guard, and the reachability test in unit-tests said so before this entry did. "
+      + "Nothing was compressed to make room and no unrelated file was touched.",
   },
   {
     key: "modules", from: 237, to: 239, date: "2026-08-17",
--- a/script/gap-tests.ts
+++ b/script/gap-tests.ts
@@ -507,20 +507,14 @@ test("messy intake: pure feeling does not force food log", () => {
   assert.equal(r.hasFoodReport, false);
 });
 
-test("isMessyLifeTranscript: short branded meal preserved whole", () => {
-  const { isMessyLifeTranscript } = UTILS;
-  assert.equal(
-    isMessyLifeTranscript("I had a McDonald's breakfast with a mocha"),
-    true,
-  );
-});
-
-test("isMessyLifeTranscript: food + feeling is messy life", () => {
-  const { isMessyLifeTranscript } = UTILS;
-  assert.equal(
-    isMessyLifeTranscript("I ate takeaways again and I feel like giving up"),
-    true,
-  );
+// isMessyLifeTranscript graded here: two life signals in one note meant "do not summarise this".
+// Cut 3 removed the summariser, so no note is summarised and the predicate has no caller. The
+// promise is inverted onto the pipeline, where it holds for notes the predicate never matched.
+test("a messy-life note reaches the handlers whole, whatever signals it carries", () => {
+  const media = readFileSync("server/handlers/media.ts", "utf-8");
+  assert.match(media, /const forBrain = transcribedText;/,
+    "every note is routed whole — no predicate decides which ones are safe to shorten");
+  assert.ok(!/condenseVoiceRamble/.test(media), "and there is nothing left to shorten them with");
 });
 
 
--- a/script/pg-acceptance-runner.ts
+++ b/script/pg-acceptance-runner.ts
@@ -228,6 +228,19 @@ export const ACCEPTANCES: Acceptance[] = [
     // opposite-defect controls. A crash or an absent verdict is itself red.
     command: ["bash", "script/red-on-revert-cut2-meal-slot.sh"] },
 
+  { id: "long-voice-tail", title: "A long note reaches the handlers whole",
+    // CUT 3 (2026-09-12). The cleaner sent 1,500 characters to a model and returned the answer as
+    // THE TRANSCRIPT, deleting 908 characters of a three-minute note — both questions among them.
+    // Needs a real database and the real front door: the claim is about what the handlers were
+    // GIVEN, durably, and about the numbers that reached their owners afterwards.
+    command: ["npx", "tsx", "script/pg-long-voice-tail-acceptance.ts"] },
+
+  { id: "long-voice-reverts", title: "Every silent cut in the voice pipeline turns a grader red",
+    // Four truncations in one pipeline — the cleaner's window, its missing lower bound, the
+    // condenser, and the ledger's own cap — plus three opposite-defect controls. Two graders,
+    // because the acceptance drives the TEXT door where the cleaner never runs.
+    command: ["bash", "script/red-on-revert-cut3-long-voice.sh"] },
+
   { id: "journey-lab", title: "Six critical journeys through the real system",
     // THE SIX JOURNEYS (#170). Same database, same migrations, same front door — a second job
     // would be a second copy of this infrastructure for no gain. It is in this runner for the same
--- a/script/pg-long-voice-tail-acceptance.ts
+++ b/script/pg-long-voice-tail-acceptance.ts
@@ -186,12 +186,15 @@ REAL("\n3. 8,500 IS STILL 8,500 — the number is not reshaped on the way to its
 // ══════════════════════════════════════════════════════════════════════════════════════════════
 REAL("\n4. NOTHING IS INVENTED FROM A NOTE THIS LONG (Cut 2 must stay true here)");
 // ══════════════════════════════════════════════════════════════════════════════════════════════
+// THE MEAL-SLOT CLAIM THAT USED TO STAND HERE IS REMOVED, not relaxed. It asserted "no meal slot
+// appears that the client did not name" and checked only that each label belonged to the
+// breakfast/lunch/dinner/snack enum — which is also satisfied by ZERO meal rows, and this note
+// produces zero. It read as attribution proof and measured nothing of the kind.
+//
+// Cut 2's own acceptance grades attribution non-vacuously, on rows that exist, at three pinned
+// hours, with nine red-on-revert cases behind it. Restating it here weakly could only ever
+// weaken it.
 {
-  const meals = (await pool.query<{ meal_label: string | null }>(
-    "SELECT meal_label FROM meal_logs WHERE user_id = $1", [user.id])).rows;
-  chk(meals.every(m => m.meal_label === null || ["breakfast", "lunch", "dinner", "snack"].includes(m.meal_label!)),
-    "no meal slot appears that the client did not name",
-    `labels=${JSON.stringify(meals.map(m => m.meal_label))}`);
   const bodies = await wire();
   chk(bodies.length > 0, "the client got an answer to a three-minute note", `${bodies.length} bodies`);
   chk(!/8\.5|85 steps|850 steps/.test(bodies.join("\n")),
@@ -216,9 +219,11 @@ REAL("\n5. WHAT THIS CUT DOES NOT FIX, RECORDED RATHER THAN CLAIMED");
   chk(alone.length === 1 && alone[0].meal_label === "lunch",
     "the same sentence ALONE logs correctly, as the lunch they called it",
     `rows=${JSON.stringify(alone)}`);
-  REAL(`    NOTE — inside the long note that sentence produced ${inNote} meal row(s). Not asserted`);
-  REAL(`    either way: first-match-wins gives one turn to one owner, which predates this cut and`);
-  REAL(`    is Cut 6's subject. Cut 3's job was to stop the words being deleted before they get there.`);
+  REAL(`    OUTSTANDING, UNDER CUT 6 — inside the long note that sentence produced ${inNote} meal`);
+  REAL(`    row(s), and the reply addresses neither of the client's two questions. Not asserted`);
+  REAL(`    either way: first-match-wins gives one turn to one owner, which predates this cut.`);
+  REAL(`    Cut 3 stops the words being deleted before they get there; Cut 6 must make the coach`);
+  REAL(`    act on the account it now receives whole.`);
 }
 
 // ══════════════════════════════════════════════════════════════════════════════════════════════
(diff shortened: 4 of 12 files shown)


ButiMM:
**CI has now run, and it confirms what the PR body predicted.** Recording it once rather than leaving the red unexplained.

| job | result |
|---|---|
| `test` (tsc + `npm test` 38/38 + gauntlet) | **success** |
| `unit-baseline` | **success** |
| `pg-acceptance` | **failure** — at "Every acceptance, isolated, on real PostgreSQL" |

The failing acceptance is `pg-information-value-acceptance`, 4 checks, all of them about a **weekend** question:

```
FAIL  the real customer front door delivers the authorized weekend question
FAIL  the reactive question becomes durable only after it is included in the handed-off reply
FAIL  'I was travelling and didn't track' durably closes the open question
FAIL  'it was normal' is accepted as the bounded answer to the question actually open
```

**Not this PR's.** Before writing any of this I checked out `017efd9` — unmodified main, this PR's own base — into a separate git worktree and ran that acceptance there. All four fail identically, with none of this cut's code present.

It was green in CI on Cut 2's head yesterday, **Friday 11 September**. It is now **Saturday 12 September**. An acceptance about the weekend, green on weekdays and red on the weekend: the same clock-dependence class as #240, which was green for a year of afternoons because nobody ran it before midday.

**No fix exists yet and I have not written one here.** It needs its own reproduction, its own regrade and its own red-on-revert — the shape Cut 1B took — and folding it into a voice-pipeline cut would widen this PR and bury a real defect inside an unrelated diff. Recorded as the next ordered infrastructure subcut.

Nothing in this cut's own evidence is affected: the new `long-voice-tail` and `long-voice-reverts` entries both ran in that same job before the red, and `tsc`, the 38-suite run and the 7/7 red-on-revert are all green on this exact head.

---

ButiMM:
## Amended — head `3d5cca48ce6fbb888ee83248475fdf7206f7ec16`

All three defects were real. Fixed here, none of them worked around.

### 1. The cleaner still silently deleted content

Reproduced your case: a stubbed reply carrying 60% of the cleaned section, **588 characters gone** including a workout correction and a question, accepted under both finish reasons. The 50% floor and word-overlap test cannot tell a faithful clean from a faithful **prefix**.

Completeness is now proven by three gates, each failing closed:

| gate | refuses |
|---|---|
| `finish_reason !== "stop"` | a reply the model did not finish — **and an absent reason**, because that is not proof of completion either |
| floor `0.8` | a reply that lost a fifth of the head |
| `coversTheEnd` | a prefix that never reaches the head's last words |

**Each gate is graded alone**, with a fixture only it can catch — a 90% prefix (end-coverage), a reply that keeps the ending but loses half the middle (floor), and a complete faithful repair reported as unfinished (finish_reason). Otherwise two could be deleted while the section stayed green on the third.

The correction and the question sit **inside the cleaned head**, asserted through `splitForClean` rather than assumed. The spelling repair the cleaner exists for is kept as a positive control — *samp*, not *stamp*, still lands.

### 2. The harness could falsely claim success

It counted any non-zero exit as "caught". A detection is now a **graded failed assertion**: the grader must reach its own verdict line, that verdict must report failures, and at least one `FAIL` must be printed. A missing verdict is a crash and fails the harness *even when the other grader legitimately went red*. Both graders must also pass **unmodified first**:

```
CONTROL: both graders are GREEN unmodified — detections below are real.
1  the window truncates again            voice-provenance-tests
2  the length floor removed              acceptance + voice-provenance-tests
2b the finish-reason gate removed        voice-provenance-tests
2c the end-coverage gate removed         voice-provenance-tests
3  the condenser returns                 acceptance + voice-provenance-tests
4  the ledger cap returns                acceptance
5  the split drops the tail              voice-provenance-tests
6  CONTROL: the cleaner stops cleaning   voice-provenance-tests
7  CONTROL: the record mangles numbers   acceptance
RED-ON-REVERT: GREEN — 9/9 cases caught.
```

### 3. The assertion that claimed more than it measured

Removed, not relaxed. Cut 2's acceptance grades attribution non-vacuously, on rows that exist, at three pinned hours, with nine revert cases behind it. The missing meal and unanswered questions now print as **OUTSTANDING, UNDER CUT 6** rather than as a neutral note.

### Three fixture faults of my own, found while writing the above

Recorded because each was the same class as the defect under repair: facts that drifted into the tail they were meant to avoid; a positive control echoing my *guess* at the head instead of the head the splitter actually sends; and a finish_reason case whose reply equalled its input, so acceptance and refusal produced the same string and it passed with the gate deleted.

### The exact final customer body, measured on this head

```
Lerato — 8,000 steps is your target. No steps logged yet today — send your count:
"8,500 steps" or "walked 5km".
```

That is the Cut 6 gap in one line: their 8,500 is stored, against yesterday, and the coach asks for it anyway while answering neither question. Cut 3 makes the words arrive; it does not make the coach act on them.

### Gates

```
tsc --noEmit          clean
npm test              38/38 green, none skipped
red-on-revert         9/9, with the unmodified control
pg-acceptance         35 run, 35 green, 0 red  — on the COMBINED head (this + #245)
```

CI on this PR stays red until **#245** merges; that is the weekend-fixture repair, built separately as instructed, with no production policy changed.

---

ButiMM:
CTO review of amended head 3d5cca48ce6fbb888ee83248475fdf7206f7ec16 — HOLD: correction/question can still be deleted inside the head.

#245 is now merged as ca1a85c161b8e0d796dc23e9b24cb783fa1f7329. I independently verified actual Tests run 34687400681: test, unit-baseline and pg-acceptance all succeeded on exact head 06e2e6009be6175db78efd25af914377c15ab0cf. Railway reports SUCCESS on the merge SHA, startup ready, DB reachable. The weekend blocker is cleared; no workflow-dispatch workaround is needed.

Accepted amendments: incomplete 60% prefixes are refused; unmodified controls and graded assertion failures replace crash-as-detection; the vacuous meal-slot assertion is removed. These are real improvements.

Remaining P1 in server/understanding/sa-transcript.ts:
I executed the fetched cleaner deterministically with a stubbed model response, NOT production audio and NOT a live model.
Fixture:
```js
const correction = "Actually I missed my Tuesday workout. What should I do today?";
const raw =
  "I went to work and took the taxi like every other day. ".repeat(18)
  + correction + " "
  + "I came home and made tea and told you about my day. ".repeat(12)
  + " FINAL7788.";
const { head, tail } = splitForClean(raw);
const modelReply = head.replace(correction, "");
// Return modelReply with finish_reason: "stop".
```
Raw: 1687 chars; head: 1467 chars; correction verified inside head.
The response preserves 95.84% of head and the full ending. All current gates accept it.
Returned: 1626 chars; the workout correction AND the question are gone; tail survives.
Therefore finish_reason + 0.8 floor + ending-word overlap do not establish a whole transcript.

Adjudicated continuation — same PR, same owner, no new model/handler/composer:
1. The cleaner is authorized for narrowly vetted local spelling repairs only, not clause rewriting. Replace the permissive set-overlap acceptance with an ordered token/span edit check. Accept case/spacing/punctuation changes and explicitly approved local spelling repairs (retain a valid stamp-to-samp control); preserve lexical token order and repetition. Unexplained deletion, insertion or substitution must return the WHOLE raw transcript. Numbers, days, negation, corrections and questions must not be changed or silently dropped. Do not solve this by raising the length percentage again or adding an AI judge.
2. Add the exact middle-deletion case above, plus deletion/alteration of a negation and a spoken/numeric quantity; the valid spelling repair and unchanged transcript must still pass. Run behavioral red-on-revert on the edit-validation gate, with passing unmodified controls; crashes are invalid evidence.
3. Retain the completed fixes: suffix carry-through, condenser removal, provenance and refusal of unfinished model responses.
4. Bring the branch up to fresh main ca1a85c, update this PR's stale body (it still names the previous head and 7/7), run focused tests first, then the full gates once on the final head, and push the precise SHA. Do not remove or loosen the failing preservation assertions.
5. Do not absorb Cut 6 into Cut 3. The measured long-note response still ignores the two questions and misses food while storing yesterday's steps; that customer failure remains explicit and queued under Cut 6. "No steps logged yet today" must not be labelled a false claim about yesterday's stored steps.

No separate architecture memo or new issue. I will inspect the amended diff and exact-head checks before merging.

ButiMM:
CTO final-head review — 70ed20cf8a22d75b5c69222ea560219a31b779f1.

Independently verified Tests run 34693570948: pg-acceptance, test and unit-baseline all SUCCESS. The deleted-middle-clause case is fixed. HOLD on two reproducible truth holes in onlyApprovedRepairs; not a CI wait.

I executed the exact added pure helpers (type annotations removed, no model call):
onlyApprovedRepairs("I have pain", "I have pap") === true
onlyApprovedRepairs("I am sad", "I am pap") === true
onlyApprovedRepairs("My change was -5 kg", "My change was 5 kg") === true
The valid stamp -> samp repair passes, and deleting the correction now correctly fails.

Decision, no further product adjudication required:
1. Replace target-word membership + edit distance <=3 with explicitly vetted FROM->TO spelling pairs. Approve stamp->samp for this cut; do not authorize arbitrary near-neighbours of a food/slang word. For any unlisted change, return the WHOLE raw transcript. Removing editDistance/target guessing is preferable to growing a vocabulary programme.
2. Preserve signed quantities and decimal values through punctuation normalization. "-5" must not become "5"; "8.5" must not become "85" or "8,500". Case/spacing and harmless punctuation repairs can remain, but punctuation is not globally semantically free. Do not silently discard non-ASCII lexical content in this check.
3. Add the exact pain/mood/sign cases above plus decimal and non-ASCII lexical-loss controls. Keep stamp->samp and unchanged transcript as positives. Extend the existing behavior grader and mutation checks, passing unmodified first; do not count a crash as detection.
4. Focused check first; then full gates once on the amended final head, update PR description, push. Same PR and existing module. No new review memo. I own final inspection/merge/deployment; don't wait for a founder decision on these repairs.

CUT 4 SCOPE APPROVED after Cut 3 merges:
A small pure shared transcript-sanity predicate in the existing voice module is permitted; it is not a new response owner. Route EVERY provider/retry result through the existing admission floor before cleaning/coaching/durable writes. Keep actual provider quality metrics where available; when unavailable, use narrow deterministic checks for empty/no-speech output and pathological repetition, not guessed confidence or English fluency. Valid SA words, slang, code-switching and numbers must pass. Retain the existing refusal/clarification response; no new fallback mouth. Cover Scribe and all Whisper attempts, including forced-English retry. Prove rejected results don't reach handleMessage or write facts, and valid results do; use stub-driven execution and label it honestly, not live STT accuracy. The source trace identifies a bypass, not measured transcription WER. First inspect provider metadata rather than assuming none exists.

Use remaining credits on this bounded amendment, then Cut 4 if capacity remains. If handing off, push/record exact SHA, clean/dirty state, focused evidence and remaining work; Codex continues that same cut, not a competing implementation. Cut 6's measured food/question failure stays queued, not erased or absorbed here.

ButiMM:
CTO MERGED Cut 3 at f8094567d2d5099dae3b10ee85038e373919223d, exact approved head 95221747f5ba11f22c9e5eb74ffcd440d70d7719. Verified all three CI jobs and PG logs: 35/35 green, long-voice mutations 11/11 with passing controls. Independently executed final pure edit helpers: pain/mood/sign/decimal/non-ASCII-loss rejected; stamp->samp accepted. Existing long-voice routing/food/question failure remains Cut 6, and live audio replay remains outstanding.

Claude: fetch fresh origin/main and START the already-approved Cut 4 now. No additional founder approval or review memo is required.

The provider-metadata question is settled from official documentation, not guessed:
https://elevenlabs.io/docs/api-reference/speech-to-text/convert
The scribe_v2 response example explicitly includes words[].logprob and language_probability.
Retain text plus provider/available word logprob metadata through the EXISTING scribeTranscribe return; update its existing caller(s). language_probability is language-identification confidence, NOT transcript correctness; do not use it as an accuracy floor. Do not claim Scribe logprob is calibrated identically to Whisper's avg_logprob or copy a rejection threshold without supporting evidence.

Whisper retries should preserve the available verbose_json segment metrics just as attempt 1 does. Ensure every provider/attempt, including Scribe and forced-English retry, reaches the same admission boundary before cleaning/coaching/fact writes. Retain existing quality policy where metrics support it; where usable metrics are absent, the approved narrow deterministic sanity predicate covers empty/no-speech and pathological repetition. Do not classify SA slang, valid one-word speech or code-switching as garble merely for not being fluent English. No new transcriber, handler, prompt authority, fallback mouth or AI judge.

Stub-driven behavioral tests must exercise all four paths, prove rejects do not reach the coach or write facts, and preserve valid SA/code-switched notes. Source-only checks are supplementary. Keep the existing clarification response. Label this quality-floor coverage, not measured WER or production-audio proof. Focused tests first, full gate once on the final candidate, then exact SHA and evidence.

Codex takes over the SAME cut when Claude stops; do not start a competing implementation. Next product-response job remains Cut 6: record the long note's food/correction and answer its questions coherently through the existing owners.

PR #245, opened 2026-09-12
The information-value acceptance is deterministic on every day of the week

ButiMM:
Base `017efd9d38e39c38a59d9247d7c58ad533ea6dd2`. Head `06e2e6009be6175db78efd25af914377c15ab0cf`.

**The CI unblocker for #244.** Test-only: no runtime file is touched, and no production policy is changed to satisfy a fixture.

## The defect, and it is not in the product

`pg-information-value-acceptance` was green every weekday and red every weekend. Nobody knew, because it had never run on a Saturday — it ran on one today and blocked #244, a cut that touches none of this code.

Measured on **unmodified main `017efd9`** in a separate worktree, Saturday 12 September — the same four checks that passed in CI the previous afternoon:

```
FAIL  the real customer front door delivers the authorized weekend question
FAIL  the reactive question becomes durable only after it is included in the reply
FAIL  'I was travelling and didn't track' durably closes the open question
FAIL  'it was normal' is accepted as the bounded answer to the question actually open
```

**The mechanism.** The fixture seeds meals on weekdays only, so the weekend is the unknown under test, then sends `"Dinner is rice, mince and mixed veggies"` through the real front door. That logs a meal for **today**. On a Saturday, today *is* a weekend day — so the live turn manufactures the very evidence the section exists to find missing, the INVESTIGATE owner correctly stops asking for a weekend it can now see, and four checks fail against correct behaviour.

## The repair

The day is **chosen rather than inherited**: everything runs on a fixed Wednesday, which makes every `ago(n)` offset, every seeded row and every live turn deterministic at every hour of every day. `Date.now()` alone is not enough — `new Date()` reads the system clock directly and never consults it — and the pin is installed *before* the server modules load, because they capture the clock as they come up. Same pattern as #240, one axis over: that one was the hour, this is the day.

**All four original durability assertions are kept verbatim.** Nothing is relaxed; they now run on every day instead of five in seven.

## And the weekend is now graded, which it never was

Section E pins a **Saturday** and a **Sunday** and asserts the promise the old fixture could only ever fail:

```
PASS  the Saturday fixture seeds weekdays only
PASS  on a Saturday with weekday-only evidence the weekend is still the missing fact
PASS  a meal logged ON a weekend day IS weekend evidence — this is why the day matters
PASS  …so the coach stops asking for the weekend it can now see
PASS  the Sunday fixture seeds weekdays only
PASS  on a Sunday with weekday-only evidence the weekend is still the missing fact
PASS  …and a Sunday log closes it for the same reason a Saturday one does
```

Both days, because being right on Saturday and never running on Sunday is the same defect one day wide.

**Section E had exactly that bug on its first run**, and it is kept as revert case 2: it reused the Wednesday-derived offsets, seeded a meal on the Saturday itself, and graded nothing. The offsets are derived from the pinned day now.

## red-on-revert 3/3

```
CONTROL  the unmodified acceptance is GREEN — detections below are real

1  the whole file pinned to that Saturday    RED — the original 4 failures, on demand
2  section E reuses the Wednesday offsets    RED — 1
3  CONTROL: weekendLoggedDays always says    RED — 16; the suite still grades the
   the weekend is covered                          product, not the calendar
```

**The harness runs an unmodified control first.** Without it, every case can report "caught" while the acceptance was failing for a reason unrelated to the mutation — a dead database, a bad import, a timeout. Detection means nothing unless the same command passes untouched. A missing verdict (a crash) fails the harness rather than counting as a detection.

## Gates

```
tsc --noEmit          clean
npm test              38/38 green
pg-acceptance-runner  33 run, 33 green, 0 red
red-on-revert         3/3, with the control
governors             unmoved — no runtime file touched
```

## Changed files

```
M  script/pg-information-value-acceptance.ts
M  script/pg-acceptance-runner.ts          (registers the revert script)
A  script/red-on-revert-information-value-clock.sh
```

---

### DIFF
--- a/script/pg-acceptance-runner.ts
+++ b/script/pg-acceptance-runner.ts
@@ -117,6 +117,12 @@ export const ACCEPTANCES: Acceptance[] = [
     // later backdated answer are all real rows; this grades the one canonical INVESTIGATE owner.
     command: ["npx", "tsx", "script/pg-information-value-acceptance.ts"] },
 
+  { id: "information-value-clock", title: "The information-value acceptance is deterministic on every day",
+    // It was green every weekday and red every weekend, and it blocked PR #244 — a cut that
+    // touches none of this code — on a Saturday. Case 1 reproduces those four failures on demand
+    // by pinning the file to that Saturday, so the defect is provable on a Tuesday.
+    command: ["bash", "script/red-on-revert-information-value-clock.sh"] },
+
   { id: "behaviour-patterns", title: "Evidence-backed behavioural patterns",
     // BEHAVIOURAL PATTERN STATE (#217). Repetition, attributed outcomes, user isolation and
     // contradiction are properties of longitudinal rows. The profile must then be read by the
--- a/script/pg-information-value-acceptance.ts
+++ b/script/pg-information-value-acceptance.ts
@@ -17,6 +17,48 @@ process.env.TWILIO_AUTH_TOKEN = "test";
 process.env.TWILIO_WHATSAPP_NUMBER = "+27000000000";
 process.env.NODE_ENV = "production";
 
+// ══════════════════════════════════════════════════════════════════════════════════════════════
+// THE CLOCK IS PINNED, AND THAT IS THE WHOLE SUBCUT (2026-09-12).
+// ══════════════════════════════════════════════════════════════════════════════════════════════
+//
+// THIS ACCEPTANCE WAS GREEN ON WEEKDAYS AND RED ON THE WEEKEND. Measured on unmodified
+// 017efd9 on Saturday 12 September 2026 — four checks fail, on a build that passed the same four
+// in CI the previous afternoon:
+//
+//     FAIL  the real customer front door delivers the authorized weekend question
+//     FAIL  the reactive question becomes durable only after it is included in the reply
+//     FAIL  'I was travelling and didn't track' durably closes the open question
+//     FAIL  'it was normal' is accepted as the bounded answer to the question actually open
+//
+// THE MECHANISM, not a guess: the fixture seeds meals on WEEKDAYS ONLY so the weekend is the
+// unknown under test, then sends "Dinner is rice, mince and mixed veggies" through the real front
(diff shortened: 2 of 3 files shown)


ButiMM:
## CI has not run on this PR, and I cannot start it

**No `Tests` workflow run exists for head `06e2e6009be6175db78efd25af914377c15ab0cf`.** This PR was opened at 10:02 UTC; #244's run fired normally at 10:19, so the `pull_request` trigger is working generally — this head simply never got a run. I tried `workflow_dispatch` and got `403 Resource not accessible by integration`, so I have no means to start one. Saying so once rather than leaving the absence to be read as a pass.

**Correcting myself:** I earlier described this PR as "CI green". That was wrong. I read it off `mergeable_state: clean`, which only means there is no merge conflict — it says nothing about checks. There are no checks.

### What is verified, and how

| evidence | result |
|---|---|
| `tsc --noEmit` on this head, locally | clean |
| `npm test` on this head, locally | 38/38 green |
| `pg-acceptance-runner` on this head, local real PostgreSQL | **33 run, 33 green, 0 red** |
| `red-on-revert-information-value-clock.sh` | 3/3, with the unmodified control |
| the same runner on this head **merged with #244's amended head** | **35 run, 35 green, 0 red** |

### Independent confirmation that the defect is main's, not #244's

main's **own nightly scheduled `Tests` run** fired this morning at 06:35 UTC on `017efd9` — no PR, no branch, nobody's changes:

```
run 855 · schedule · main · 017efd9
  test           success
  pg-acceptance  FAILURE   at "Every acceptance, isolated, on real PostgreSQL"
  unit-baseline  skipped   (PR-only by design)
```

Identical job signature to #244's. **main is red today**, on a Saturday, with nothing merged into it since Cut 2. That is this PR's diagnosis confirmed by a run I did not create and cannot influence.

The worktree reproduction on unmodified `017efd9` in the PR body, the failing nightly on main, and the local green on this head are three independent lines pointing at the same thing: a fixture that inherited the day of the week.

I will keep watching this PR. If a run appears I will report it; I cannot make one appear.

---

PR #246, opened 2026-09-12
Cut 4: every STT result reaches the admission floor

ButiMM:
## Cut 4 final candidate

Base: `f8094567d2d5099dae3b10ee85038e373919223d`

Head: `0798243b89adfded275fc9bedb4bab7b30df9e06`

This keeps the existing voice architecture and makes every STT result pass the shared admission floor while preserving provider provenance.

### Behavior

- Raw provider transcript is retained unchanged in voice provenance.
- Only recognized no-speech markers are removed from handler/customer text.
- Arbitrary bracketed client speech is preserved.
- Mixed `[BLANK_AUDIO]` plus real speech reaches handlers as only the real speech; the marker cannot reach stored interpreted text or the outbound echo.
- Whisper metrics remain provider-scoped and unchanged.
- Eight-identical-word runs remain refused.
- The blanket `vocabulary.size <= 2` rejection is removed; low-vocabulary client speech is admitted.
- Open-loop, missed-session and session-owner PostgreSQL fixtures use deterministic test clocks only. No coaching policy changed.

### Current files

- `server/elevenlabs.ts`
- `server/handlers/media.ts`
- `server/understanding/sa-transcript.ts`
- `script/pg-acceptance-runner.ts`
- `script/pg-missed-session-outbound-acceptance.ts`
- `script/pg-open-coaching-loop-acceptance.ts`
- `script/pg-session-owner-acceptance.ts`
- `script/pg-stt-admission-acceptance.ts`
- `script/red-on-revert-cut4-stt-admission.sh`
- `script/red-on-revert-voice-provenance.sh`
- `script/voice-provenance-tests.ts`

### Final blocker and correction

CI run `34823374866` was 36/37 PostgreSQL acceptances. The sole red was `session-owner`; its first failed assertion was:

```
...and advances the programme cursor, which is what owns 'which session is next'
{"before":2,"after":2}
```

The fixture started at programme day 2 but inherited Monday. Monday maps calendar slot 1 back to next day 2, so the stored cursor value correctly remained 2 and the date-sensitive assertion misread that as no advance. The only follow-up change pins this test-only fixture to Wednesday, the matching day-2 slot.

### Evidence

Focused real-PostgreSQL acceptance:

```
pg-session-owner-acceptance: GREEN - all checks passed
```

The complete local PostgreSQL inventory executed once on Windows/PostgreSQL 18. `session-owner`, all Cut 4 voice suites, open-loop and missed-session were green in sequence. The aggregate local result was 23/37 because Windows child-process and PostgreSQL 18 baseline differences affect unrelated suites; no unrelated code was changed.

Authoritative Ubuntu/PostgreSQL 16 CI on exact head `0798243b89adfded275fc9bedb4bab7b30df9e06`:

- Run: https://github.com/ButiMM/KamLife-Coach/actions/runs/34826894140
- `unit-baseline`: SUCCESS
- `test`: SUCCESS
- `pg-acceptance`: SUCCESS
- PostgreSQL summary: `37 run, 37 green, 0 red`
- `session-owner`: GREEN
- `open-coaching-loop`: GREEN
- `missed-session-outbound`: GREEN
- `voice-provenance` and reverts: GREEN
- `stt-admission` and reverts: GREEN

No merge performed. CTO owns inspection and merge.

### DIFF
--- a/script/pg-acceptance-runner.ts
+++ b/script/pg-acceptance-runner.ts
@@ -247,6 +247,18 @@ export const ACCEPTANCES: Acceptance[] = [
     // because the acceptance drives the TEXT door where the cleaner never runs.
     command: ["bash", "script/red-on-revert-cut3-long-voice.sh"] },
 
+  { id: "stt-admission", title: "A garbled transcript writes nothing, a real one writes everything",
+    // CUT 4 (2026-09-12). The garble floor read `if (voiceQuality && …)` and voiceQuality is set
+    // only by Whisper attempt 1 — so Scribe (which runs FIRST in production), the catch retry and
+    // the forced-English retry all SKIPPED it. Needs a real database because the claim is about
+    // what a refused transcript does NOT store, and what an admitted one does.
+    command: ["npx", "tsx", "script/pg-stt-admission-acceptance.ts"] },
+
+  { id: "stt-admission-reverts", title: "Every bypass of the admission floor turns a grader red",
+    // One mechanism per case: the metrics-only condition returning, each content check removed,
+    // and two opposite-defect controls so a floor that refuses everything cannot pass.
+    command: ["bash", "script/red-on-revert-cut4-stt-admission.sh"] },
+
   { id: "journey-lab", title: "Six critical journeys through the real system",
     // THE SIX JOURNEYS (#170). Same database, same migrations, same front door — a second job
     // would be a second copy of this infrastructure for no gain. It is in this runner for the same
--- a/script/pg-stt-admission-acceptance.ts
+++ b/script/pg-stt-admission-acceptance.ts
@@ -1,0 +1,239 @@
+/**
+ * REAL-POSTGRESQL ACCEPTANCE — a transcript that fails the admission floor writes nothing
+ * durable, and one that passes it writes everything (Cut 4, 2026-09-12).
+ *
+ * ══════════════════════════════════════════════════════════════════════════════════════════════
+ * WHAT WAS TRUE BEFORE THIS CUT, reproduced on f809456 before a line of it was written
+ * ══════════════════════════════════════════════════════════════════════════════════════════════
+ *
+ * The garble floor in media.ts read:
+ *
+ *     if (voiceQuality && wordCount >= 2 && (avgLogprob < -1.0 || comp > 2.5))
+ *
+ * and `voiceQuality` is populated ONLY by Whisper attempt 1, the one call that asks for
+ * `response_format: "verbose_json"`. Scribe — which runs FIRST whenever ELEVENLABS_API_KEY is set,
+ * so it is the production path — the catch retry, and the forced-English retry all leave it null.
+ * `voiceQuality &&` then SKIPS the floor rather than meeting a weaker one.
+ *
+ * Driven through the real branch on that SHA, sixteen consecutive "you" reached handleMessage and
+ * came back to the client as «🎤 I heard: "you you you…"». The floor was not failing. It was not
+ * running.
+ *
+ * ══════════════════════════════════════════════════════════════════════════════════════════════
+ * WHAT THIS GRADES, AND HOW HONESTLY
+ * ══════════════════════════════════════════════════════════════════════════════════════════════
+ *
+ * STUB-DRIVEN EXECUTION of the real voice branch against a REAL database. The audio download and
+ * the two STT clients are stubs; everything else — the SSRF guard, the length caps, the word-count
+ * guard, the admission floor, the cleaner, the refusal floor, the handler recursion and every
+ * durable writer behind it — is the shipped code running against PostgreSQL.
+ *
+ * THIS SAYS NOTHING ABOUT TRANSCRIPTION ACCURACY. It grades which STRINGS are admitted and what
+ * they do to the database. No claim here rests on real-world WER, and none is made.
+ *
+ * A CORRECTION TO A STANDING NOTE, recorded rather than quietly fixed: both
+ * script/pg-voice-provenance-acceptance.ts and script/pg-long-voice-tail-acceptance.ts say the
+ * voice branch "cannot be executed offline" because assertSafeMediaUrl demands an https URL on an
+ * allow-listed PUBLIC host. That was true of the fixtures those cuts had, and it is not a property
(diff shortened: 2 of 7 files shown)


ButiMM:
CTO consolidated review of ffdb49726acdf812018f9107f0e806808a387630 — HOLD pending the following bounded amendment. PR opened by CTO; do not wait for founder permission to open/continue it.

Accepted: shared predicate at the existing refusal boundary; stub-driven real voice execution, preserved SSRF guard; real PG rejection checks; unmodified controls and graded failures rather than crashes. Actual audio accuracy/WER remains unmeasured. Cut 3 is independently confirmed deployed at f809456.

Four items, one amendment, same Cut 4:

1. COMPLETE THE ALREADY-APPROVED METADATA WORK.
server/elevenlabs.ts still narrows Scribe to text. Official API response includes words[].logprob:
https://elevenlabs.io/docs/api-reference/speech-to-text/convert
Widen the existing scribeTranscribe return and update existing callers so text and available provider-tagged word logprobs survive together. language_probability identifies language; it is not transcript-correctness confidence. Preserve/log metadata, but do not pretend Scribe scores share Whisper calibration or invent a new rejection threshold. Missing metadata uses the narrow deterministic floor.
Whisper attempts 2 and 3 should also request verbose_json and collect actual finite segment metrics, not zero-fill absent values. Every accepted result owns its OWN metrics. Clear/replace attempt-1 quality when using a later result. Source shows an empty attempt-1 result can set bad metrics, then a valid forced-English result inherits them; add the exact behavioral control: empty attempt 1 with bad segments -> good attempt 3 with good segments must reach coaching.

2. REMOVE THE UNAPPROVED FREQUENCY REJECTION.
I executed the exact new predicate without any model call:
transcriptFailsAdmission("pap pap eggs pap pap chicken pap pap beans pap pap fish", null) === true.
This contains multiple different foods and no eight-token run; rejecting it because a staple exceeds 60% is broader than the approved loop/no-speech policy. Delete commonest-token >0.6 as a rejection criterion. Retain narrowly specified consecutive/alternating-loop checks, with admitted repetitive but meaningful food/number/SA cases. Do not expand to English-fluency or low-vocabulary judgements masquerading as confidence.

3. COVER ACTUAL NO-SPEECH MARKERS THROUGH THE BRANCH.
The predicate admits "[BLANK_AUDIO] [BLANK_AUDIO]" today (executed exact helper). These contain letters, so the no-letter/no-digit check cannot identify them despite comments claiming it does. Recognize only explicit no-speech markers, not arbitrary bracketed SA text. Exercise repeated markers through the real voice branch: existing clarification, no handler recursion, no meal/step writes. Keep existing supported single-word command behavior unchanged; a pure-helper test is not proof that the full branch admits every one-word food/feeling. Do not claim universal one-word support in this cut.

4. UPDATE THE EXECUTED HARNESS, NOT JUST SOURCE STRINGS.
Changing retry formats will invalidate stubs that treat EVERY verbose_json request as attempt 1. Track invocation/attempt explicitly. Prove:
- valid-looking transcript with bad Whisper metrics is refused on attempts 1/2/3;
- good later metrics are not poisoned by an earlier empty result;
- Scribe words/logprobs reach the shared boundary while valid configured Scribe text remains admitted;
- explicit no-speech/loops reject before handler/fact writes;
- valid SA/code-switch/quantities/repetitive meal list still pass;
- valid controls and each required mutation are behaviorally graded. Crashes remain invalid evidence.
Focused checks first, then full gate once on the final head; update PR body and push exact SHA. Report local results separately from CI.

Claude can complete this if capacity remains; otherwise Codex continues this SAME branch/PR immediately, with the pushed head as the handoff. CTO owns inspection/merge/deployment. No parallel builder, new issue, or additional memo. Cut 6 remains the next named coaching-response failure; no claim here that admitted speech is automatically fully understood.

ButiMM:
CTO review of amended head dbf9963bcc8ce3631bb8da2b3f2dab114a7cbfc6 — one same-cut blocker remains.

Accepted and independently inspected:
- Scribe parser now returns finite words[].logprob plus language_probability, provider-tagged; Scribe values are observable but are not judged by Whisper thresholds.
- all three Whisper attempts request verbose_json;
- take() replaces text and metrics together, including null for absent metrics;
- bad empty attempt-1 metrics cannot poison the forced-English result;
- the >60% frequency rule is deleted;
- explicit marker-only output and SA/code-switch/quantity controls exist;
- exact-head CI run 34709143368: test, unit-baseline and pg-acceptance all SUCCESS.

BLOCKER — the mixed-marker claim is false in the shipped path:
transcriptFailsAdmission() computes
  const spoken = stripNoSpeechMarkers(t)
only inside its boolean decision. It never returns spoken and media.ts never assigns it. After admission, media.ts sends the ORIGINAL transcribedText to cleanSATranscript, turnVoice, handleMessage and the client echo.
The test for "[BLANK_AUDIO] " + GOOD asserts only that /samp and beans/ survived. It never asserts that BLANK_AUDIO disappeared. So it passes while the marker still reaches the coach/customer. This is the same evidence gap the cut is meant to remove.

Bounded amendment; no product decision needed:
1. Expose the existing exact-whitelist transformation under a clear name (for example stripKnownNoSpeechMarkers) and reuse it in both the predicate and the existing media path. Do not duplicate the marker vocabulary.
2. Preserve provenance: retain the provider's exact raw transcript in the existing voice trace/provenance field. Hand only the marker-stripped candidate to word-count, cleaner, handler and echo. If stripping leaves no speech, take the existing unable-to-transcribe/clarification path; no new response wording.
3. The mixed fixture must assert:
   - reachedHandlers equals/contains GOOD and contains no BLANK_AUDIO;
   - final client body contains no BLANK_AUDIO;
   - marker + real speech still reaches the existing facts/response path;
   - raw provider provenance still contains the original marker (use the existing voice provenance mechanism, not a new table/field).
4. Add one behavioral red-on-revert case that removes only the media-path sanitation while leaving predicate sanitation intact. It must turn a grader red; a source-string check alone is insufficient.
5. Retain arbitrary non-whitelisted bracketed content unchanged. Marker-only, mixed-marker, Scribe metadata, retry metrics, repetitive-food and SA-language controls all remain green.
6. Focused tests, then full gate once on final head; push exact SHA and update PR body. Do not open another PR or wait for founder confirmation.

This is the final known Cut 4 blocker. After it passes, CTO inspects and merges. Cut 6 remains next; no additional review memo or diagnosis.

